PR #3, opened 2021-10-29
Export endpoint

sarahmcdougall:
# Summary
These changes expose a system level $export endpoint to the export server and enable polling logic with a `bulkstatus` endpoint.

## New behavior
A system level $export endpoint has been created, which exports data from a FHIR server, whether or not it is associated with a patient (as opposed to the patient endpoint or group of patients endpoint). The $export endpoint supports the optional query string params `_outputFormat` and `_type`. (The `_outputFormat` parameter is the format for the requested bulk data files to be generated and defaults to `”application/fhir+ndjson”`. The `_type` parameter is a string of comma-delimited FHIR resource types.) If successful, the endpoint responds with `202 Accepted` status and generates a unique polling location endpoint in the `content-location` header. If unsuccessful, an appropriate error is thrown in the server. 

An endpoint was added for the bulk status requests, and it responds with the status that is in the database. This follows similar logic to the bulk status endpoint for bulk import in `deqm-test-server`.

## Code changes
`bulkstatus.service.js` was added to check the status of the bulk export request. `export.service.js` was added to execute the bulk export and to validate the export parameters (`_outputFormat` and `_type`). Note that currently `bulkExport` only adds the pending bulk export request, validates the parameters ,and then replies with status code 202 if successful, which populates the `Content-location` header.

Prettier was added to the project.

# Testing Guidance
Run the unit tests and ensure that they pass. To test on your local machine, 
Run the command `node src/server/server.js` or `npm start` to spin up the server. Then navigate to Insomnia and test out the `$export` and `bulkstatus` endpoints and check that the parameter validation for `$export`. Here are some possible test cases for `$export`:

* Send a GET request to `http://localhost:3000/$export` and check that the response body is empty, the response status is `202 Accepted`, and the `content-location` header is populated with a client ID to be used for the `bulkstatus` endpoint.
* Send a GET request to `http://localhost:3000/$export?_outputFormat=application/fhir+ndjson&_type=Encounter,Condition` and check that you get the same result (empty response, `202 Accepted`, populated `content-location`).
* Send a GET request with the `_outputFormat` set to `application/ndjson` and `ndjson`. Both should work the same as the previous example.
* Send a GET request with an unsupported `_outputFormat` (ex. `_outputFormat=unsupported`) and ensure that you get a `400` response status and error message.
* Send a GET request with an unsupported `_type` (ex. `_type=unsupported`) and ensure that you get a `400` response status and error message.
* Send a GET request with the `_since` parameter and ensure that you get a `400` response status and error message since we do not currently support it.
* Send a GET request with any other parameter(s) and ensure that you get a `400` response status and error message listing the unrecognized parameter(s)

Here is a possible test for `bulkstatus`:
* Send a GET request to `http://localhost:3000/$export` and copy the `id` in the `content-location` header. Paste it to the end of the base url and send a GET request (`http://localhost:3000/bulkstatus/<id>`. You should get a `202 Accepted` response code with the `x-progress` and `retry-after` headers populated. The status will not change from “In Progress” at this time, but you can try hardcoding the status and checking the other cases outlined in `bulkstatus.service.js`.


hossenlopp (changes requested):
Various suggestions for improvements.

Biggest issue is that it created a bulk data status tracker in the DB anyways when the parameters fail. I suggest taking the a TDD approach when fixing it. Create or add checks to an existing test that compare the count of DB items before and after a bad parameter request to make sure it didn't change. After seeing this new/updated failing test it can be fixed and a test showing the fix works will already be there.

sarahmcdougall:
Thank you for the detailed suggestions @hossenlopp :) Please let me know if I missed/misinterpreted anything in regard to my changes.

hossenlopp (approved):
Looks like a good start for this little server.

PR #5, opened 2021-11-03
Export to ndjson

MegRiley:
# Summary
Added the ability to export all supported resources to ndjson files. 

## New behavior
Added support for exporting the resources from the database into ndjson files. 
The function will either export all the resources on the system OR if the request contains a _type parameter it will only export the resources specified in that parameter.
The files should have a name pattern that is similar to this './tmp/Patient/patient123456.ndjson'  except that patient should be replace with the specified resource type either from the _type parameter or from the supported resources list.  

## Code changes
exportToNDJson - new file that contains which exports all resources of all resource types to ndjson  or only the resources specified in the request.query.type parameter 
# Testing guidance
Since this isn't hooked into the server directly I tested it by running it as a script, I followed the same pattern as in the createConnectathonBundle script. (Also this can be seen in the early revs of the file) Adding a main function and then adding the call to main at the end of the file.   Once executed this should create a ./tmp/ folder and under that it should create a system of sub directories one for each resource. The files should have a name pattern that is similar to this './tmp/Patient/patient123456.ndjson'  except that patient should be replace with the specified resource type either from the _type parameter or from the supported resources list.


src/util/exportToNDJson.js
```diff
@@ -0,0 +1,41 @@
+const { db } = require('./mongo');
+const supportedResources = require('../util/supportedResources');
+const fs = require('fs');
+
```
natjoe4, line 4: Nit pick, but this function could use a jsdoc comment.

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,41 @@
...
+const exportToNDJson = async (clientId, request) => {
+  let dirpath = './tmp/';
+  await fs.promises.mkdir(dirpath, { recursive: true });
+  //create list of requested types if request.query._types param exists
+  if (request.query._types) {
```
natjoe4, line 9:
```suggestion
  if (request.query._type) {
```

natjoe4 (changes requested):
I get `TypeError: supportedResources.foreach is not a function` when running the added function in isolation. The test file also wasn't named using the #{module}.test.js pattern, which prevented the added tests from being run. When I renamed it I got `TypeError: createTestResource is not a function`.

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+const exportToNDJson = async (clientId, request) => {
+  let dirpath = './tmp/';
+  await fs.promises.mkdir(dirpath, { recursive: true });
+  let requestTypes;
+  if (request.query._type) {
+    requestTypes = request.query._type.split(','); //this is the list types to export
+  } else {
+    //create list of requested types if request.query._type param doesn't exist
+    requestTypes.push(supportedResources);
```
lmd59, line 19: I'm a little confused why this is a push if it's initially creating the requestTypes list

lmd59: Could potentially do this whole block as a `||` with "optional chaining" (safe operator) to avoid the empty requestTypes declaration?
something like `requestTypes = request.query._type?.split(',') || supportedResources`
or could use a ternary operator

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  let lineCount = 0;
+  requestTypes.foreach(type => {
+    dirpath = dirpath + type.toString();
```
lmd59, line 24:
```suggestion
    dirpath += type.toString();
```
(just for cleaner line)

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,35 @@
...
+    let collections = db.collection(type);
+    const filename = dirpath + '/' + type + clientId + '.ndjson';
```
lmd59, line 26:
```suggestion
    const filename = `${dirpath}/${type}_${clientId}.ndjson`
```
(just for cleaner line)

lmd59: Just as a double check, is there a reason that we use the type in the filepath twice? Doesn't super matter either way.

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,35 @@
...
+    collections.foreach(function (doc) {
+      let result = JSON.parse(JSON.stringify(doc));
+      fs.appendFileSync(dirpath, (++lineCount === 1 ? '' : '\r\n') + JSON.stringify(result));
```
lmd59, line 29: Is there something that disallows a newline at the end of the file? Also not sure why we're stringifying then parsing then stringifying again. If not disallowed, I would probably just
```suggestion
      fs.appendFileSync(dirpath, JSON.stringify(doc) + '\n');
```
Then you can ditch lines 22 and 28 as well.

test/exportToNDJson.test.js
```diff
@@ -0,0 +1,38 @@
...
+    _type: 'Patient'
+  }
+};
+const mockRequestWithoutType = {
+  id: 'mockRequestWithoutType',
+  query: {}
+};
+const clientId = '123456';
+//= ./tmp/ + type.toString(); + resourceType + clientId + '.ndjson';
+const expectedFileName = './tmp/Patient/patient123456.ndjson';
```
lmd59, line 18:
```suggestion
const expectedFileName = './tmp/Patient/patient_123456.ndjson';
```
...if you add in my suggested `_` above

test/exportToNDJson.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+  beforeEach(async () => {
+    await app.ready();
+  });
+
+  test('Expect folder created and export successful  when _type  parameter is present', async () => {
+    exportToNDJson(clientId, mockRequestWithType);
+    expect(fs.existsSync(expectedFileName)).toBe(true);
+  });
+  test('Expect folder created and export successful  when _type  parameter is  not present', async () => {
```
lmd59, line 30:
```suggestion
  test('Expect folder created and export successful when _type  parameter is not present', async () => {
```
Just me being a Word processor at you :-P

test/exportToNDJson.test.js
```diff
@@ -0,0 +1,38 @@
...
+//= ./tmp/ + type.toString(); + resourceType + clientId + '.ndjson';
+const expectedFileName = './tmp/Patient/patient123456.ndjson';
+describe('check export logic', () => {
+  beforeAll(createTestResource(testPatient, 'Patient'));
+
+  beforeEach(async () => {
+    await app.ready();
+  });
+
+  test('Expect folder created and export successful  when _type  parameter is present', async () => {
```
lmd59, line 26:
```suggestion
  test('Expect folder created and export successful when _type  parameter is present', async () => {
```
Just me being a Word processor at you :-P

lmd59 (commented):
Also- are there any negative tests that make sense? Like any times we wouldn't expect a folder to be created? Are there any errors for misformatted request that we want to handle?

lmd59:
we gotta figure out this unit test issue too!

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,57 @@
...
+  fs.promises.mkdir(dirpath, { recursive: true });
+  let filename = dirpath + type + clientId + '.ndjson';
+
+  let lineCount = 0;
+  fs.open(filename, 'w', function (err) {
+    if (err) throw err;
+  });
+  if (doc) {
+    doc.forEach(function (doc) {
+      var stream = fs.createWriteStream(filename, { flags: 'a' });
```
mgramigna, line 48:
```suggestion
      const stream = fs.createWriteStream(filename, { flags: 'a' });
```

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,57 @@
...
+
+      let result = JSON.parse(JSON.stringify(doc));
```
mgramigna, line 50:
```suggestion
      const result = JSON.parse(JSON.stringify(doc));
```

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,57 @@
...
+const getDocuments = async (db, collectionName) => {
+  const query = {};
+  let doc = await db.collection(collectionName.toString()).find(query).toArray();
+  return { document: doc, collectionName: collectionName.toString() };
+};
+
+const writeToFile = async (doc, type, clientId) => {
+  let dirpath = './tmp/';
+  fs.promises.mkdir(dirpath, { recursive: true });
+  let filename = dirpath + type + clientId + '.ndjson';
```
mgramigna, line 40:
```suggestion
  const filename = path.join(dirpath, `${type}.ndjson`);
```

I'd prefer if the clientId was in a subdirectory name, not the filename itself. That makes the file names too hard to read. You will need to import the `path` module from node as well to implement this suggestion.

This will give us `tmp/client-id/Patient.ndjson` instead of `tmp/Patientclient-id.ndjson`

mgramigna (changes requested):
I gave this a test too to try to help out and I found some things that are causing the functionality to fail for me. I think most of my inline comments should fix some stuff.

In addition, this code needs to be called from the kickoff endpoint so we start writing the NDJson after the HTTP request is sent. You can just add a function call after [this line](https://github.com/projecttacoma/bulk-export-server/blob/da68798c08d1d6ab72b70fc6703f4c0fc5faa1ae/src/services/export.service.js#L12). **Important**: Don't use `await`. We want to kick off this function synchronously so the endpoint still responds while the bulk data stuff happens in the background

test/exportToNDJson.test.js
```diff
@@ -0,0 +1,41 @@
...
+  query: {
+    _type: 'Patient'
+  }
+};
+const mockRequestWithoutType = {
+  id: 'mockRequestWithoutType',
+  query: {}
+};
+
+const expectedFileName = './tmp/123456/patient.ndjson';
```
mgramigna, line 17:
```suggestion
const expectedFileName = './tmp/123456/Patient.ndjson';
```

pretty sure it will take the resource type directly so you might run into a capitalization issue here

mgramigna: actually looks like this doesn't cause an issue lol. I guess fs not case sensitive

mgramigna (approved):

mgramigna (approved):

src/util/exportToNDJson.js
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const writeToFile = async (doc, type, clientId) => {
+  let dirpath = './tmp/' + clientId;
+  fs.mkdirSync(dirpath, { recursive: true });
+  const filename = path.join(dirpath, `${type}.ndjson`);
+
+  let lineCount = 0;
+  fs.writeFileSync(filename, '');
+  fs.open(filename, 'w', function (err) {
```
natjoe4, line 45: This is working on my machine without creating the file with fs.writeFileSync or fs.open before we create the stream

natjoe4 (approved):

PR #6, opened 2021-11-10
Export status responds with ndjson

sarahmcdougall:
# Summary
Logic was added to the export status endpoint to respond with the ndjson file locations, and for GET requests sent to these locations to respond with the actual ndjson content.

## New behavior
Now, when a GET request is sent to `http://localhost:3000/$export` to export data from a FHIR server and write it to ndjson files, the server responds with status endpoint 200 OK, and responds with the ndjson URLs for all the desired resources. When a GET request is sent to one of these locations, the actual ndjson content is retrieved from the file that was saved to disk, and the server responds with the actual ndjson content. 

## Code changes
The ndjson retrieval endpoint was added to `app.js`. The `bulkstatus.service.js` file was updated so that the ndjson urls are retrieved once the bulk status is complete. A new file, `ndjson.service.js`, was created to retrieve the actual ndjson contents when the user sends a GET request to a given url. Errors are thrown if the location cannot be found.

# Testing guidance
A good way to test this is to run `npm run db:reset` followed by `npm run connectathon-upload`. Then send a GET request to $export, and copy the contents of the `content-locatio` header. Use this location and send a GET request to it (ex. `http://localhost:3000/bulkstatus/<your-id>`). In the response headers, the `expires` header should be populated and the response body should contain an object containing an array of all the ndjson urls. Then, you can send a GET request to the client id (ex. `http://localhost:3000/<your-id>`) and the response body should contain the actual ndjson content.

Run the unit tests and ensure that they pass. Please let me know if I should use a different approach to unit testing - I hardcoded the `tmp` directory with client `12346` but there's probably a cleaner approach that I could use. For the unit test for ndjson content retrieval, Jest couldn't read the test ndjson file that I tried to bring in, so I instead just checked that the response body is defined (not that it equals a specific thing). Let me know if you don't think this is sufficient.

### DIFF
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -1,5 +1,9 @@
 const { exportToNDJson } = require('../util/exportToNDJson');
-const { addPendingBulkExportRequest } = require('../util/mongo.controller');
+const {
+  addPendingBulkExportRequest,
+  updateBulkExportStatus,
+  BULKSTATUS_COMPLETED
+} = require('../util/mongo.controller');
 const supportedResources = require('../util/supportedResources');
 
 /**
@@ -13,6 +17,7 @@ const bulkExport = async (request, reply) => {
     const clientEntry = await addPendingBulkExportRequest();
     exportToNDJson(clientEntry, request);
     reply.code(202).header('Content-location', `/bulkstatus/${clientEntry}`).send();
+    updateBulkExportStatus(clientEntry, BULKSTATUS_COMPLETED);
   }
 };
 
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -11,7 +11,7 @@ const bulkExport = async (request, reply) => {
   if (validateExportParams(request, reply)) {
     request.log.info('Base >>> $export');
     const clientEntry = await addPendingBulkExportRequest();
-    await exportToNDJson(clientEntry, request, reply);
+    exportToNDJson(clientEntry, request);
     reply.code(202).header('Content-location', `/bulkstatus/${clientEntry}`).send();
   }
 };
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -11,7 +11,6 @@ const bulkExport = async (request, reply) => {
   if (validateExportParams(request, reply)) {
     request.log.info('Base >>> $export');
     const clientEntry = await addPendingBulkExportRequest();
-    console.log('about to export');
     await exportToNDJson(clientEntry, request, reply);
     reply.code(202).header('Content-location', `/bulkstatus/${clientEntry}`).send();
   }
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -1,9 +1,5 @@
 const { exportToNDJson } = require('../util/exportToNDJson');
-const {
-  addPendingBulkExportRequest,
-  updateBulkExportStatus,
-  BULKSTATUS_COMPLETED
-} = require('../util/mongo.controller');
+const { addPendingBulkExportRequest } = require('../util/mongo.controller');
 const supportedResources = require('../util/supportedResources');
 
 /**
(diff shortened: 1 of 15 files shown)


src/services/export.service.js
```diff
@@ -13,6 +17,7 @@ const bulkExport = async (request, reply) => {
     const clientEntry = await addPendingBulkExportRequest();
     exportToNDJson(clientEntry, request);
     reply.code(202).header('Content-location', `/bulkstatus/${clientEntry}`).send();
+    updateBulkExportStatus(clientEntry, BULKSTATUS_COMPLETED);
```
mgramigna, line 20: I don't think this function call will behave as we want it to in practice. This is a function for marking the bulk export job as completed, which should only happen after all the ndjson is exported. Right now, this is kicked off synchronously after the initial response is sent that the export has started.

I think we should move this function call to within the `exportToNDJson` function, and call it with `await` after all the `writeToFile`s have finished.

src/util/mongo.controller.js
```diff
@@ -120,6 +120,15 @@ const getBulkExportStatus = async clientId => {
   return status;
 };
 
+/**
+ * Wrapper for the updateResource function that updates the bulk status
+ * @param {*} clientId The id signifying the bulk status request
+ * @param {*} newStatus
+ */
+const updateBulkExportStatus = async (clientId, newStatus) => {
+  updateResource(clientId, { status: newStatus }, 'bulkExportStatuses');
```
mgramigna, line 129:
```suggestion
  await updateResource(clientId, { status: newStatus }, 'bulkExportStatuses');
```

I think we need this since we do care about waiting until the DB operation finished before returning from here

src/services/ndjson.service.js
```diff
@@ -0,0 +1,24 @@
...
+const returnNDJsonContent = async (request, reply) => {
+  request.log.info('Base >>> return NDJSON');
+  const clientId = request.params.clientId;
+  const url = request.params.fileName;
+  const filePath = `tmp/${clientId}/${url}`;
+  if (fs.existsSync(filePath)) {
+    const readStream = fs.createReadStream(`tmp/${clientId}/${url}`);
+    readStream.on('data', function (text) {
+      reply.header('Content-type', 'application/ndjson+fhir').send(text);
+    });
```
mgramigna, line 18: This current pattern logs a fastify error to the console:

image

I think this is telling us that the function sends a void response before processing the stream. I read online that you can send the stream directly back with the reply rather than using the `on('data')` hook and it will work exactly the same, but this error no longer happens.

```suggestion
    const readStream = fs.createReadStream(`tmp/${clientId}/${url}`);
    reply.header('Content-type', 'application/ndjson+fhir').send(readStream);
```

src/services/export.service.js
```diff
@@ -11,7 +11,7 @@ const bulkExport = async (request, reply) => {
   if (validateExportParams(request, reply)) {
     request.log.info('Base >>> $export');
     const clientEntry = await addPendingBulkExportRequest();
-    exportToNDJson(clientEntry, request);
+    await exportToNDJson(clientEntry, request, reply);
```
mgramigna, line 14: we don't actually want to `await` this. We intentionally left it out so it would kick off in the background and not block returning from the initial kickoff request.

```suggestion
    exportToNDJson(clientEntry, request, reply);
```

test/bulkstatus.service.test.js
```diff
@@ -19,13 +28,16 @@ describe('checkBulkStatus logic', () => {
...
     await supertest(app.server)
-      .get('/bulkstatus/COMPLETED_REQUEST')
+      .get(`/bulkstatus/${clientId}`)
       .expect(200)
       .then(response => {
         expect(response.headers.expires).toBeDefined();
         expect(response.headers['content-type']).toEqual('application/json; charset=utf-8');
-        expect(response.body).toBeDefined();
+        expect(response.body.outcome).toEqual([
+          { type: 'Patient.ndjson', url: 'http://localhost:3000/testClient/Patient.ndjson' }
```
mgramigna, line 39:
```suggestion
          { type: 'Patient.ndjson', url: `http://localhost:3000/${clientId}/Patient.ndjson` }
```

mgramigna (changes requested):

mgramigna (approved):

PR #7, opened 2021-11-15
Update bulk status to respond with FHIR type not file name

mgramigna:
When I was reviewing this code, I missed that the `type` part of the returned export data should just be a fhir resource type, not the file name. This PR strips off the `.ndjson` extension and updates the test properly.

To test, run an export and then hit the bulkstatus endpoint after it completes. The `type`s should be FHIR types, not file names now.


natjoe4 (approved):
Works for me 👍

PR #8, opened 2021-11-18
Operation outcome

natjoe4:
# Summary
Bulk Export Server now returns an OperationOutcome object on failed bulk export.

## New behavior
Previously, the Bulk Export Server would throw an error on $export fail. Now, it returns an OperationOutcome resource with error information

Added functionality for catching smaller errors to turn into warnings and return to the user on a successful bulk export.

## Code changes

- Added src/util/errorUtils and the createOperationOutcome function which is a utility to convert errors into operation outcomes
- Made updates to exportToNDJson function to integrate the createOperationOutcome function
- Added utility functions and documentation to support eventually catching warnings that will not cause the bulk export to fail but should still be returned to the user
- Updated testing to reflect the new expected output

# Testing guidance
- Run all unit tests
- Try throwing an error at any point during the `exportToNDJson` function in `src/util/exportToNDJson.js`. This should force the export operation to fail and return an OperationOutcome with the message you passed into the error

### DIFF
--- a/src/util/errorUtils.js
+++ b/src/util/errorUtils.js
@@ -1,7 +1,7 @@
 const { v4: uuidv4 } = require('uuid');
 
 /**
- *
+ * Takes in error information and create and OperationOutcome to return to the user
  * @param {*} message a message to be displayed inside the OperationOutcome
  * @param {*} param1 an object containing the issueCode and severity to be copited into the OperationOutcome
  * @returns An OperationOutcome to return to the user with information about the failure
--- a/src/util/errorUtils.js
+++ b/src/util/errorUtils.js
@@ -1,0 +1,34 @@
+const { v4: uuidv4 } = require('uuid');
+
+/**
+ *
+ * @param {*} message a message to be displayed inside the OperationOutcome
+ * @param {*} param1 an object containing the issueCode and severity to be copited into the OperationOutcome
+ * @returns An OperationOutcome to return to the user with information about the failure
+ */
+
+const { htmlEncode } = require('htmlencode');
+
+function createOperationOutcome(
+  message,
+  {
+    issueCode = 'processing', // http://hl7.org/fhir/valueset-issue-type.html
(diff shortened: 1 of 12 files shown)


src/util/errorUtils.js
```diff
@@ -0,0 +1,32 @@
+const { v4: uuidv4 } = require('uuid');
+
+/**
+ * Takes in error information and create and OperationOutcome to return to the user
```
sarahmcdougall, line 4:
```suggestion
 * Takes in error information and create an OperationOutcome to return to the user
```

src/util/errorUtils.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {*} message a message to be displayed inside the OperationOutcome
+ * @param {*} param1 an object containing the issueCode and severity to be copited into the OperationOutcome
```
sarahmcdougall, line 6:
```suggestion
 * @param {*} param1 an object containing the issueCode and severity to be copied into the OperationOutcome
```

src/util/mongo.controller.js
```diff
@@ -123,10 +121,24 @@ const getBulkExportStatus = async clientId => {
...
+  if (error) {
+    await updateResource(clientId, { status: newStatus, error: error }, 'bulkExportStatuses');
+  } else {
+    await updateResource(clientId, { status: newStatus }, 'bulkExportStatuses');
+  }
+};
+
+/**
+ * Adds a warning to the bulkstatus warning array
+ * @param {*} clientId the client id for the request which
```
sarahmcdougall, line 136: should it be "the client id for the request which produces a warning"?

sarahmcdougall (approved):
Super cool that this is getting added to the export server. I just have a few suggestions for comments but that's it! I think the tests are very thorough, and both functionality and documentation are great. I was able to throw an error in the `exportToNDJSON` function and see the OperationOutcome resource get returned in the response.

PR #9, opened 2021-11-29
updated spec discrepancies

natjoe4:
# Summary
Updated a minor discrepancies between our implementation of bulk export and the HL7 IG

## New behavior
- The url found in the `content-location` header of an $export request response is now an absolute url rather than a relative url
- The results of a successful bulkstatus requests after completing an export are now stored on `response.data.output` instead of `response.data.outcome`

## Code changes
- Made above two changes
- Updated testing

# Testing guidance
- Run the unit tests
- Boot up the server and send a `GET` request to `http://localhost:3000/$export`
- Ensure the `content-location` response header is an absolute url (ex. `http://localhost:3000/bulkstatus/{CLIENT_ID}` instead of `bulkstatus/{CLIENT_ID}`
- Send a `GET` request to the `content-location` header url and ensure that the array storing the paths to the NDJson files is under the `output` property instead of the `outcome` property

### DIFF
--- a/src/services/bulkstatus.service.js
+++ b/src/services/bulkstatus.service.js
@@ -22,7 +22,7 @@ async function checkBulkStatus(request, reply) {
     reply.send({
       transactionTime: new Date(),
       requiresAccessToken: false,
-      outcome: responseData,
+      output: responseData,
       // When we eventually catch warnings, this will add them to the response object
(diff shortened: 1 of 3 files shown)


sarahmcdougall (approved):
flawless

PR #11, opened 2021-12-07
Bee queue in export server

sarahmcdougall:
# Summary
This PR makes use of the [bee queue](https://github.com/bee-queue/bee-queue) Node.js queue library in the bulk export server so that we can adopt a publish/subscribe model for our bulk data kickoff requests.

## New behavior
When a request is sent to the endpoint that handles the bulk data kickoff, a new event is created for the kickoff request. This new event gets enqueued onto the bee-queue. Once enqueued, the listener “subscribed” to the event will complete the bulk data logic under the hood - it will query Mongo for the data to export and write it to ndjson format.

This approach makes use of [Redis](https://redis.com), an in-memory data storage solution.

## Code changes
A Redis GitHub Actions configuration was added for Redis. In `export.service.js`, the queue is brought in and creates a job for the kickoff request. The actual bee-queue is constructed in `exportQueue.js`, which specifies the queue options and the `process()` function for handling the jobs.

# Testing guidance
To get Redis set up on your local machine, run: `brew install redis`, `brew services start redis`, and then `redis-cli ping` to check that the Redis server is running. This is outlined in the updated README.
Install bee-queue with `npm install bee-queue`.
Run the unit tests and ensure that they pass. Note that the unit tests use Redis and do not try to mock out a Redis server. There are no explicit tests to check the queueing functionality - rather, the successful tests in `export.service.test.js` check that the job spy has been called. I ran into an issue where some of the unit tests create open handles in Jest when the queue is not closed immediately after the test. For now, the `afterEach` for the tests closes the queues so that we do not have the open handles. In the future, we will need to investigate whether separate queues are being created within some of the tests and how we can solve that issue so that there is a single queue.

To test this yourself, run `npm run start` to start up the server, then spam the server with `$export` requests. Check that no errors come up and that you can access the bulk status for the jobs that are created.


src/resources/exportQueue.js
```diff
@@ -0,0 +1,24 @@
...
+// Create a new queue to establish new Redis connection
+const exportQueue = new Queue('export', queueOptions);
+
+// This handler pulls down the jobs on Redis to handle
+exportQueue.process(async job => {
+  // Payload of createJob exists on job.data
+  const { clientEntry, types } = job.data;
+  // Call the existing export ndjson function that writes the files
+  const d = await exportToNDJson(clientEntry, types);
+  return d;
```
mgramigna, line 17: `exportToNDJson` returns `void` right? I think this will always return `undefined`. We can probably just `await` the function withour returning anything and it'll still be fine

mgramigna (changes requested):
Code looks good and works. Just one small comment

mgramigna (approved):

PR #12, opened 2021-12-10
Exclude non patients

MegRiley:
# Summary
For the test data that we use as a reference in the bulk-export-server, we upload all the bundles from the connectathon repo. These bundles also include Libraries, Measures,  ValueSets, and MeasureReports. These resourceTypes aren't important for testing bulk-import since they aren't patient data, so for now we can just filter those resourceTypes out of the bundles that are uploaded as part of the connectathon bundle script.
## New behavior
The script will no long upload bundles to the database that are not of resource type "Patient"
## Code changes
 Added a check in the connectathonBundle.js to check that the resource type is Patient. I wrapped the `createResource` call in this check to prevent the upload of the other bundles that aren't relevant to bulk import behavior
# Testing guidance
Pull the connectathon repo to the root of the project and execute the bundle upload script. when you look at the database (I used robo mongo) you shouldn't see any bundles that aren't patient specific bundles. (unless your database had previously been in use and had data from a prior time.


natjoe4 (changes requested):
According to the task description, we want to keep everything but Libraries, Measures,  ValueSets, and MeasureReports. I believe this code will only upload the actual patient resources, but will leave out Encounters, Conditions, Observations, etc... I was unable to populate the DB with anything but actual patient resources.

natjoe4 (approved):
Looks good to me!

PR #13, opened 2022-01-04
Child Process Functionality

natjoe4:
# Summary
Updated the Bee-Queue configuration to spin up child processes at server startup to handle bulk data requests

## New behavior
Hopefully none while interacting with the server, with the exception of a couple new logs, but all bulk data requests will now be handled in child processes preventing the server from being blocked by processing.

## Code changes
- Added ExportWorker.js file with all code to process bulk export requests
- Updated server.js to spin up

# Testing guidance
- Start up the server with `PORT=3001 npm run start`
- You should now see `export-worker-{PID}: Export Worker Started!` log twice
- Try spamming export requests to `http://localhost:3001/$export` (Ensure you have already reset the db and run a connectathon upload)
- You should see `export-worker-{PID}: Processing Request: {CLIENT_ID}` although it may get buried in the logs.
- Upon completion of the request, you should see `export-worker-{PID}: Completed Export Request: {CLIENT_ID}`
- Try sending a request to the bulkstatus endpoint to ensure that the request suceeds
- Now fire up the deqm-test-server using `npm run start`
- Kick off a `$import` request by POSTing a fhir parameters object with the exportUrl `http://localhost:3001/$export`
- Query the bulkstatus endpoint returned and ensure that the request suceeded
- Finally, power down both servers, then run `EXPORT_WORKERS=20 npm run start` in the bulk export server and ensure you get a warning about starting more processes than the number of available cpus.

### DIFF
--- a/src/resources/exportQueue.js
+++ b/src/resources/exportQueue.js
@@ -1,20 +1,14 @@
+// Setup for import queue which pushes jobs to Redis
+
 const Queue = require('bee-queue');
-const { exportToNDJson } = require('../util/exportToNDJson');
 
 const queueOptions = {
-  removeOnSuccess: true
+  removeOnSuccess: true,
+  isWorker: false
 };
 
 // Create a new queue to establish new Redis connection
 const exportQueue = new Queue('export', queueOptions);
-
-// This handler pulls down the jobs on Redis to handle
-exportQueue.process(async job => {
-  // Payload of createJob exists on job.data
-  const { clientEntry, types } = job.data;
-  // Call the existing export ndjson function that writes the files
-  await exportToNDJson(clientEntry, types);
-});
 
 exportQueue.on('error', err => {
   console.log('queue error: ', err);
--- a/src/resources/exportQueue.js
+++ b/src/resources/exportQueue.js
@@ -1,8 +1,11 @@
 // Setup for import queue which pushes jobs to Redis
-
 const Queue = require('bee-queue');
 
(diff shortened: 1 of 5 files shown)


src/resources/exportQueue.js
```diff
@@ -1,21 +1,15 @@
+// Setup for import queue which pushes jobs to Redis
+
 const Queue = require('bee-queue');
-const { exportToNDJson } = require('../util/exportToNDJson');
 
 const queueOptions = {
-  removeOnSuccess: true
+  removeOnSuccess: true,
+  isWorker: false
```
mgramigna, line 7: should we add the same redis environment options that you added in the worker?

src/server/exportWorker.js
```diff
@@ -0,0 +1,28 @@
+const { client } = require('../util/mongo.js');
+const { exportToNDJson } = require('../util/exportToNDJson');
+const Queue = require('bee-queue');
+
+console.log(`export-worker-${process.pid}: Export Worker Started!`);
+
+const exportQueue = new Queue('export', {
+  redis: {
+    host: process.env.REDIS_HOST || 'localhost',
+    port: process.env.REDIS_PORT || 6379
```
mgramigna, line 10: these env variables could/should be specified in our `.env` file

mgramigna (changes requested):
Looks good. two small comments

mgramigna (approved):

PR #15, opened 2022-01-13
Connectathon Changes

natjoe4:
# Summary
Updated the retry-after header to be 1 second and implemented 429:TooManyRequests error

## New behavior
- Retry After is reset from 120 -> 1 
- Spamming 10 bulkstatus requests in 1 second results in 429:TooManyRequests error

## Code changes
Above changes and updated testing

# Testing guidance
You probably wont be quick enough to see the 429, so you'll need to make a couple modifications
- Add the following sleep function to line 47 of src/util/exportToNDJson.js:
- ``` await new Promise(resolve => setTimeout(resolve, 20000));```
- This will force the code to sleep for 20 seconds before completing the request
- Now, change the ```RETRY_AFTER``` constant in src/services/bulkstatus.service.js from 1 to 20.
- Fire up the test server with `npm run start`
- Send a `GET` request in Insomnia to `http://localhost:3000/$export`
- Copy the returned `content-location` header and paste it into a new `GET` request
- Ensure to receive `202:Accepted`
- Now spam the send request button and ensure that you eventually receive `429:TooManyRequests`
- Wait the required 20 seconds, then send another bulkstatus request and ensure the job has completed with `200:OK`

### DIFF
--- a/src/util/mongo.controller.js
+++ b/src/util/mongo.controller.js
@@ -101,6 +101,8 @@ const addPendingBulkExportRequest = async () => {
   const bulkExportClient = {
     id: clientId,
     status: BULKSTATUS_INPROGRESS,
+    numberOfRequests: 0,
+    timeSinceLastRequest: null,
     error: {},
     warnings: []
   };
--- a/src/util/mongo.controller.js
+++ b/src/util/mongo.controller.js
@@ -134,17 +134,28 @@ const updateBulkExportStatus = async (clientId, newStatus, error = null) => {
 };
 
 /**
- * @param {string} clientId
- * @param {Object} data
+ * Changes the first valid request time for tracking 429:TooManyRequests errors
+ * @param {string} clientId the id of the client making the export request
+ * @param {Object} timeOfFirstValidRequest a Date object storing the time of the first valid request
  */
 const updateFirstValidRequest = async (clientId, timeOfFirstValidRequest) => {
   await updateResource(clientId, { timeOfFirstValidRequest }, 'bulkExportStatuses');
 };
 
+/**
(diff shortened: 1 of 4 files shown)


src/util/mongo.controller.js
```diff
@@ -131,6 +133,33 @@ const updateBulkExportStatus = async (clientId, newStatus, error = null) => {
   }
 };
 
+/**
+ * Changes the first valid request time for tracking 429:TooManyRequests errors
+ * @param {string} clientId the id of the client making the export request
+ * @param {Object} timeOfFirstValidRequest a Date object storing the time of the first valid request
+ */
+const updateFirstValidRequest = async (clientId, timeOfFirstValidRequest) => {
```
hossenlopp, line 141: These multiple update calls are a bit of over engineering in my humble opinion. They could be single update function with arguments for the two timeout tracking fields. But this is just a nitpick and this is just our engineering test server so I wouldn't undo the work here. It makes the other logic look cleaner anyways.

test/bulkstatus.service.test.js
```diff
@@ -96,6 +96,21 @@ describe('checkBulkStatus logic', () => {
         expect(JSON.parse(response.text).message).toEqual('Could not find bulk export request with id: INVALID_ID');
       });
   });
+  test('check 429 error returned for spamming requests', async () => {
```
hossenlopp, line 99: Could use a test to make sure that after getting a 429, then appropriately waiting, you are able to get a 202 again.

hossenlopp (commented):
Comments from a pass at code review.

hossenlopp (approved):
Looks good. Does what it needs to give the proper response for the spec.

PR #16, opened 2022-02-28
Added support for _typeFilter

MegRiley:
# Summary
Added support for a_typeFilter.

## New behavior
The bulk export server can now process _typeFilters as part an export
## Code changes
- src/util/exportToNDJson.js -- added `processTypeFilter` which is called as part of `getDocuments` and is only called if the _typeFilter param was present on the request. Added logic to parse the type filter and pass it along to the value set look up code

-src/server/exportWorker.js - modified job info to now pass around the typefilter as well 
- src/services/export.service.js - modified to pull the typefilter out of the job config and pass it along to the export 
- src/util/valueSetHelper.js - a new file that I brought over 2 functions from fqm -execution `getHierarchicalCodes` and  `getCodesFromValueSet` in order to support looking up the value sets

# Testing guidance
 - run  `npm run test` ensure all tests pass
 - modify deqm-test-server (\src\server\importWorker.js)to set 
  `const output = await BulkImportWrappers.executeBulkImport(exportUrl, measureBundle, true);`

 - Send a bulk data request, that does not have _types specified.  through deqm-test-server note the number of items listed in the operation outcome when the request has been completed (I think it was around 200 for me but if you have different things in your database you may see a different number. 
 - Now send a request where the body does have _types specified. Once the request has completed you should notice that the count of imported resources is lower than the amount sent in the first case

An example request body is as follows:
`{
	"resourceType": "Parameters",
	"parameter": [
		{
			"name": "measureReport",
			"resource": {
				"resourceType": "MeasureReport",
				"measure": "http://hl7.org/fhir/us/cqfmeasures/Measure/EXM130"
			}
		},
		{
			"name": "exportUrl",
			"valueUrl": "http://localhost:3001/$export"
		}
	]
}`

Note you may need to change the port of the export server if you're not running it on 3001 
Here is an example of a bulk import request with a more explicit type filter:
  {
      "name": "_type",
      "valueString": "Patient"
    },
    {
      "name": "_type",
      "valueString": "Encounter"
    },
    {
      "name": "_type",
      "valueString": "Procedure"
    }


src/util/valueSetHelper.js
```diff
@@ -0,0 +1,41 @@
+function getHierarchicalCodes(valueSet) {
```
lmd59, line 1: My rule of thumb is typically that if a function isn't exported, then it doesn't *need* to have a jsdoc style comment on it (sometimes still nice to have). I actually am not sure this function needs to be exported since it's only used within this file (your call if you think it will be useful outside of this file). But would you mind at least adding jsdoc style comments on all exported functions?

src/util/exportToNDJson.js
```diff
@@ -92,4 +128,32 @@ const writeToFile = function (doc, type, clientId) {
   } else return;
 };
 
+const processTypeFilter = async function (typefilterLookupEntry) {
+  console.log('am I inside processTYpeFilter?');
```
lmd59: Should delete leftover console log

test/exportToNDJson.test.js
```diff
@@ -30,6 +31,10 @@ describe('check export logic', () => {
     expect(fs.existsSync(expectedFileName)).toBe(true);
   });
 
+  test('Expect folder created and export successful when _typeFilter  parameter is retrieved from request', async () => {
+    await exportToNDJson(clientId, mockType, mockTypeFilter);
+    expect(fs.existsSync(expectedFileName)).toBe(true);
+  });
```
lmd59, line 37: Can we do a test with multiple queries/resources/properties/urls as part of the filter? Could probably do multiples of all of those at once in a single filter. Seems like there's a bit of complexity with managing all of those at the same time, so it would be good to have a test for that.

src/scripts/getConnectathonBundles.js
```diff
@@ -58,7 +58,6 @@ async function main() {
           if (
             res.resource.resourceType != 'Measure' &&
             res.resource.resourceType != 'Library' &&
-            res.resource.resourceType != 'ValueSet' &&
```
lmd59, line 61: Do we want all connectathon bundle valuesets to be added? Just a double check since obviously we decided to filter valuesets out at some point, so I wasn't sure if this would cause any other problems.

src/util/exportToNDJson.js
```diff
@@ -59,12 +87,20 @@ const exportToNDJson = async (clientId, types) => {
  * @param {string} collectionName The collection of interest in the mongodb
  * @returns {Object} An object containing all data from the given collection name as well as the collection name
  */
```
lmd59, line 89: Update jsdocs

src/util/exportToNDJson.js
```diff
@@ -59,12 +87,20 @@ const exportToNDJson = async (clientId, types) => {
...
-const getDocuments = async (db, collectionName) => {
+const getDocuments = async (db, collectionName, typefilterLookup) => {
   const query = {};
```
lmd59, line 91: This is a bit of a nitpick, but I think it might be clearer to update the query const (make it a `let` or similar) with `processTypeFilter(typefilterLookup[collectionName.toString()]` (using the appropriate rules from the else case below) and then use that query object in a single `db.collection(collectionName.toString()).find(query ...` especially since the query const is just an empty `{}`

src/services/export.service.js
```diff
@@ -75,13 +76,16 @@ function validateExportParams(request, reply) {
   let unrecognizedParams = [];
   Object.keys(request.query).forEach(param => {
     if (!['_outputFormat', '_type'].includes(param)) {
```
lmd59, line 78: probably just put '_typeFilter' into this list of recognized params

src/util/exportToNDJson.js
```diff
@@ -92,4 +128,31 @@ const writeToFile = function (doc, type, clientId) {
   } else return;
 };
 
+const processTypeFilter = async function (typefilterLookupEntry) {
+  let queryArray = [];
+
+  if (typefilterLookupEntry) {
+    // throw if we don't have the value set
```
lmd59, line 135:
```suggestion
    // throw an error if we don't have the value set
```

src/util/exportToNDJson.js
```diff
@@ -92,4 +128,31 @@ const writeToFile = function (doc, type, clientId) {
...
+
+        vsResolved.forEach(code => {
+          queryArray.push({ [`${propertyValue}.coding.code`]: code.code });
+        });
+      });
+      await Promise.all(results);
+    }
+  }
+
+  let query = {
```
lmd59, line 152: Directly return this instead of doing a `let` first?

src/util/exportToNDJson.js
```diff
@@ -92,4 +128,31 @@ const writeToFile = function (doc, type, clientId) {
   } else return;
 };
 
+const processTypeFilter = async function (typefilterLookupEntry) {
```
lmd59, line 131: Might be helpful to do a jsdoc style comment here.

src/scripts/getConnectathonBundles.js
MegRiley: I think we do, because we need to be able to query mongo to determine if the code is present in the valueSet

lmd59 (approved):

PR #18, opened 2022-04-01
Support for Patient/$export

sarahmcdougall:
# Summary
Gives support for the Patient-level export endpoint in the server, using the Patient Compartment Definition as a reference for the resources that are permitted to be exported.

## New behavior
According to [the $export spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow), the All Patients endpoint obtains a detailed set of FHIR resources of diverse resource types pertaining to all patients. This endpoint supports several query parameters such as `_outputFormat`, `_since`, `_type`, and other experimental filters.

We can now send a GET request to `[base]/Patient/$export`. When doing so, the [Patient Compartment Definition](https://www.hl7.org/fhir/compartmentdefinition-patient.html)  is used as a reference for recommended resources to be returned from the operation. The operation narrows down the exportable (yes that is a word) resource types to all the resource types that could reference a Patient resource, regardless of whether or not they actually do reference any Patient resources stored on the server.

In the future, the experimental `patient` parameter (which is not currently implemented) can be used to filter the resources to export to those who belong to patients contained in the `patient` list.

## Code changes

A new route was added to the app for `/Patient/$export`. This route corresponds to `patientBulkExport()` in `export.service`, which validates the parameters, then creates a client entry, similar to the flow for system-level export.

According to the [spec](https://hl7.org/fhir/uv/bulkdata/OperationDefinition-patient-export.html), if the client asks for only resource types that are outside the Patient Compartment, the server should return an OperationOutcome in an error response to the request. This is handled with `filterPatientResourceTypes()`. 

Otherwise, if only some of the provided resource types from the `_type` parameter are included in the Patient Compartment, the array of requested types is filtered to include those that are included in the Patient Compartment. A warning is logged with the types that were filtered out of the request.

If the `_type` parameter is not present, then the `types` that are included in the enqueued job are all the resource types that appear in the Patient Compartment Definition.

The Patient Compartment Definition was brought into the project, and the `parseCompartmentDefinition` script was created to parse the Compartment Definition and compile an array of resource types that can reference Patient resources.

Unit tests were added for the new endpoint.

# Testing guidance
Test out the Patient Compartment Definition parsing script with `npm run parse-compartment-def`. This should create a file `patientExportResourceTypes` that contains an array of resource types that can reference Patient resources.

Run the unit tests for the patient-level export endpoint.

Test the system-level export endpoint (`[base]/$export`) to ensure that the endpoint’s functionality remains the same (since minor changes were made to the `types` attribute of the enqueued jobs).

Testing the new functionality in Insomnia:
* Run `npm run db:reset`, followed by `npm run connectathon-upload` to reset the database and upload the Connectathon bundles for testing.
* Test the new endpoint without the `_type` parameter. Send a GET request to `http://localhost:3000/Patient/$export`. You should receive an empty response body, `202 Accepted` status, and the `content-location` header should be populated with a URL to be used for the bulk status endpoint. Send a GET request to this URL. You should receive urls for all the resource types that are on the server and are in the Patient Compartment Definition.
* Test the new endpoint with the `_type` parameter, where one of the types is not included in the Patient Compartment Definition. To test this, I did the following:

1. Open the Mongo shell with the `mongo` command, then `use bulk-export-server` once you are in the mongo shell
2. Insert an entry for a Parameters resource. The Parameters resource is a supported resource, but it does not appear in the Patient Compartment, and so it should get filtered out. Use this command: `db.Parameters.insertOne({ "resourceType": "Parameters" })`
3. Exit the Mongo shell, start up the server, and send a GET request to `Patient/$export?_type=Condition,Parameters`. You should see a log warning about Parameters being filtered out. Send a GET request to the bulk status endpoint, and you should only see the URL for a Condition ndjson file. You can also send a GET request to `Patient/$export?_type=Condition` and you should get the same result (without any warning).

* Test the endpoint with the `_type` parameter for a resource type that does not appear in the Patient Compartment. Try sending a GET request to `Patient/$export?_type=Parameters`. You should receive an OperationOutcome since none of the requested types are allowed.
* Test the new endpoint along with the DEQM test server (send an $import request whose `exportUrl` is the bulk export server with the `/Patient/$export` endpoint). Repeat all the above tests with this setup.


test/export.service.test.js
```diff
@@ -98,3 +140,71 @@ describe('Check barebones bulk export logic (failure)', () => {
...
+    await supertest(app.server)
+      .get('/Patient/$export?_unrecognizedparam=invalid')
+      .expect(400)
+      .then(async () => {
+        expect(await db.collection('bulkExportStatuses').countDocuments({})).toEqual(numEntries);
+      });
+  });
+
+  // Close export queue that is created when processing these tests
+  // TODO: investigate why queues are leaving open handles in this file
```
natjoe4, line 205: Is this still happening? Wasn't for me.

README.md
```diff
@@ -84,6 +86,24 @@ Once MongoDB is running on your machine, run the `npm start` command to start up
...
+
+Endpoint: `GET [fhir base]/Patient/$export`
+
+#### System Level Export
+
+Export data from a FHIR server, whether or not it is associated with a patient. This supports use cases like backing up a server, or exporting terminology data by restricting the resources returned using the `_type` parameter.
+
+Endpoint: `GET [fhir base]/$export`
+
+For more information on the export endpoints, read this documentation on the [ Export Request Flow](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow).
```
natjoe4, line 105:
```suggestion
For more information on the export endpoints, read this documentation on the [Export Request Flow](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow).
```

natjoe4 (changes requested):
Functionality all looks good. Well done! I noticed, we're still exporting ValueSets here, but that may be because of the compartment definition issues I mention below. No major changes :)

test/export.service.test.js
sarahmcdougall: Now it is leaving an open handle for me after the successful tests, but I am able to remove these calls from the failure tests. I'm going to keep a `queue.close()` after the successful patient-level export tests finish so that we don't have the open handle issue.

src/services/export.service.js
```diff
@@ -89,4 +129,34 @@ function validateExportParams(request, reply) {
...
 }
 
-module.exports = { bulkExport };
+/**
+ * Checks provided types against the recommended resource types for patient-level export.
+ * Filters resource types that do not appear in the patient compartment definition and throws
+ * OperationOutcome if none of the provided types are present in the patient compartment definition.
+ * @param {Object} request http request object
+ * @param {Object} reply the response object
+ */
```
natjoe4, line 138: Missing a return on the jsdoc here.

natjoe4 (changes requested):
All looks great and works well. Just one tiny comment.

PR #19, opened 2022-04-26
Support for group export

lmd59:
# Summary
Support for Group export. This is a new endpoint on the server that uses the compartment definition to only export resources that reference patients in the Group submitted.

## New behavior
The [Group export spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#endpoint---group-of-patients), shows the new `GET [fhir base]/Group/[id]/$export` endpoint
Note from the spec that "If a FHIR server supports Group-level data export, it SHOULD support reading and searching for Group resource. This enables clients to discover available groups based on stable characteristics such as Group.identifier." ... which uh, I don't think the export server does, but our export server is pretty basic - maybe that's okay?

We can now hit this endpoint in order to export the resources associated with patients in that group (and the patient resources themselves). The [Patient Compartment Definition](https://www.hl7.org/fhir/compartmentdefinition-patient.html)  is used as a mapping for how different resources can reference a patient. The operation then only exports those resources that have a patient reference to one of the patients in the group.

## Code changes

- README and app.js updates for the new `[fhir base]/Group/[id]/$export` route.
- Added `groupBulkExport` function to the export service to handle the new endpoint, identify the patients in the group, and create an export job.
- Updates to the export worker to pass through the group patient Ids.
- Added `patientsQueryForType` function and other updates to `exportToNDJson` file to create and integrate the query object for the patient data filtering based on the group's patient ids.
- Convenience updates to mongo.controller to separate direct db calls out of the `exportToNDJson` file.
- Added tests for the export service and function tests for methods in `exportToNDJson`.
- Added files to capture the compartment definition mapping as well as new testing fixtures.

# Testing guidance
`npm run test`

Add some convenience groups to the server for testing. I added these resources directly using Robo 3T. Example Group resources:
```
{
    "id" : "EXM130-patients",
    "actual" : true,
    "member" : [ 
        {
            "entity" : {
                "reference" : "Patient/denom-EXM130"
            }
        }, 
        {
            "entity" : {
                "reference" : "Patient/numer-EXM130"
            }
        }
    ],
    "resourceType" : "Group",
    "type" : "person"
},
{
    "id" : "empty",
    "actual" : true,
    "member" : [],
    "resourceType" : "Group",
    "type" : "person"
}
```

With the above resources, you can try the endpoint, then use the content-location url returned to view the export resources. 
Request: GET http://localhost:3001/Group/EXM130-patients/$export
Expected Bulkstatus Result: Encounter, Group, Patient, and Procedure resources that reference either denom-EXM130 or numer-EXM130

Request: GET http://localhost:3001/Group/empty/$export?_type=Encounter
Expected Bulkstatus Result: Encounter resources that reference either denom-EXM130 or numer-EXM130

Request: GET http://localhost:3001/Group/empty/$export
Expected Bulkstatus Result: empty output

Feel free to futz around with the groups or filter parameters or add other resources with patient references and try out the endpoint with those.

### DIFF
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -77,6 +77,10 @@ const groupBulkExport = async (request, reply) => {
   if (validateExportParams(request, reply)) {
     request.log.info('Group >>> $export');
     const group = await findResourceById(request.params.groupId, 'Group');
+    if (!group) {
+      reply.code(404).send(new Error(`The requested group ${request.params.groupId} was not found.`));
+      return;
+    }
     const patientIds = group.member.map(m => {
       const splitRef = m.entity.reference.split('/');
       return splitRef[splitRef.length - 1];
@@ -185,7 +189,7 @@ function filterPatientResourceTypes(request, reply) {
   if (types.length !== filteredTypes.length) {
     if (filteredTypes.length === 0) {
       reply.code(400).send(
-        createOperationOutcome('None of the provided resource types are permitted for Patient-level export.', {
+        createOperationOutcome('None of the provided resource types are permitted for Patient/Group export.', {
           issueCode: 400,
           severity: 'error'
         })
@@ -193,7 +197,7 @@ function filterPatientResourceTypes(request, reply) {
     }
     const removedTypes = types.filter(type => !filteredTypes.includes(type));
     request.log.warn(
-      `The following resource types were removed from the request because they are not permitted for Patient-level export: ${removedTypes.join(
+      `The following resource types were removed from the request because they are not permitted for Patient/Group export: ${removedTypes.join(
         ', '
       )}`
     );
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -1,4 +1,4 @@
-const { addPendingBulkExportRequest } = require('../util/mongo.controller');
+const { addPendingBulkExportRequest, findResourceById } = require('../util/mongo.controller');
 const supportedResources = require('../util/supportedResources');
 const exportQueue = require('../resources/exportQueue');
 const patientResourceTypes = require('../compartment-definition/patientExportResourceTypes.json');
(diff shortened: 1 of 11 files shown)


src/services/export.service.js
```diff
@@ -67,6 +67,46 @@ const patientBulkExport = async (request, reply) => {
...
+/**
+ * Exports data from a FHIR server for resource types pertaining to patients found in the referenced
+ * Group. Uses parsed patient compartment definition similarly to patientBulkExport.
+ * @param {Object} request the request object passed in by the user
+ * @param {Object} reply the response object
+ */
+const groupBulkExport = async (request, reply) => {
+  if (validateExportParams(request, reply)) {
+    request.log.info('Group >>> $export');
+    const group = await findResourceById(request.params.groupId, 'Group');
```
sarahmcdougall, line 79: When testing, I tried sending a GET `[base]/Group/[id]/$export` request for a Group ID that does not exist on the server. I got a 500 Internal Server Error "cannot read property 'member' of null." I think what's happening is the `group` variable here becomes null and so it errors out when doing the mapping on the group members. Adding a 404 error when the group is null should prevent this 500 from being thrown

sarahmcdougall (changes requested):
Great job with this PR! I was able to test all the endpoints mentioned under Testing Guidance, and I was able to test with $import in deqm-test-server as well.

One thing I noticed when testing in deqm-test-server is that calling $import with an empty Group will throw a 500 error “Export server failed to export any resources.” I think this is fine, but not the most descriptive since it’s not clear that nothing was exported because the Group itself is empty. Maybe we can explore this more in the future if needed. 

Otherwise, I just have one in-line change and one small comment below:

- In export.service.js, we should update the messages returned by `filterPatientResourceTypes()` to say “Patient- or Group-level $export requests” instead of saying just “Patient-level $export” since these errors/warnings can be thrown when doing Group-level $export.

sarahmcdougall (approved):
Tested the new updates and everything looks good :) I ran the unit tests again and everything still passes!

PR #22, opened 2022-10-31
update DB_HOST and PORT

lmd59:
# Summary

Update issue with mongo connection client introduced with node 17 and thus affecting node updates from 16->18/lts. Also updates port to 3001 since this is typically the port this is run on in testing

## New behavior

Server connects to mongo with node 18

## Code changes
Just environment variable changes in .env to change the database host and the application port

# Testing guidance
`npm run test` and make sure functionality/tests work with node 18

### DIFF
--- a/.env
+++ b/.env
@@ -1,8 +1,8 @@
-DB_HOST = localhost
+DB_HOST = 127.0.0.1
 DB_PORT = 27017
 DB_NAME = bulk-export-server
 HOST = localhost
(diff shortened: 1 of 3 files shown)


sarahmcdougall (approved):
Tested on Node v.18.12.0 by running bulk import with the test server and everything worked as expected, was able to connect to the db using the new DB_HOST. Also tested unit tests and everything worked 👍

PR #24, opened 2022-12-01
Ecqm-content-r4-2021 upload

natjoe4:
# Summary
Updated the database population script to pull from the eqcm-content-r2-2021 repository rather than Connectathon.

## Code changes
 - `connectathon-upload` script has been changed to `upload bundles`

# Testing guidance
 - Navigate to the top level of your `bulk-export-server` directory and run: 
 - `git clone https://github.com/cqframework/ecqm-content-r4-2021.git`
 - `npm run db:reset`
 - `npm run upload-bundles`
 - The final log to terminal should read `722 resources uploaded from 15 Bundle files. 15 Groups created.`
 - Pop open Robo 3T and look at the Groups collection. Ensure the group corresponding to each measure has the correct number of members. Group ids will be the `{Corresponding Measure ID}-patients`. Take a closer look at some of these entries and make sure the referenced patients appear in the corresponding Measure Bundle.


elsaperelli (approved):
Lgtm!

PR #29, opened 2023-03-13
Add bundle upload endpoint

sarahmcdougall:
# Summary
Exposes new endpoint for POSTing transaction and batch bundles to the server. Creates script for POSTing a series of bundles in a directory.

## Motivation
When uploading transaction/batch bundles to the server, we are interested in replacing the references with new ids, which does not happen when we use the current `uploadPremadeBundles` script. Also, for small test cases where we want to upload a single transaction bundle and check export functionality with it, we want to have an easier means to upload this bundle via an endpoint.

This functionality will be used in a larger workflow that does the following:
1. Several groups of patients of different sizes are generated in Synthea, and each set of patient data is represented as a separate transaction bundle. Additionally, as outlined in the [Synthea wiki](https://github.com/synthetichealth/synthea/wiki/FHIR-Transaction-Bundles), batch bundles get created for the Practitioner resources (`output/fhir/practitionerinformation…json`) and the Organization and Location resources (`output/fhir/hospitalinformation…json`). 
2. The batch bundles for hospital and practitioner info are uploaded to the server.
3. The transaction bundles are uploaded to the server, and a Group resource is created for each group of patients using the new ids that are generated during the transaction bundle upload process.
4. The groups are used to conduct group `$export` requests from a bulk export client.

## New behavior
A user can POST a transaction of batch bundle to `http://localhost:3001/` Note the difference between batch and transaction bundles: if there is an issue with uploading the transaction bundle (i.e. one operation fails), then everything fails. With a batch bundle, if one operation fails, the rest of the operations can still be completed. See ["Batch Processing Rules"](https://www.hl7.org/fhir/http.html#brules) for more information.

## Code changes
* New endpoint in `src/server/app.js` for bundle upload
* New service `src/services/bundle.service.js` for uploading the bundle, replacing the references in the bundle, and sending back a transaction-response/batch-response bundle.
* New util file `src/util/bundleUtils.js` for replacing references (logic ported over from deqm-test-server)
* Moved `createPatientGroupsPerMeasure` so it can be used between multiple scripts
* New script `postTransactionBundles.js`
    * Based on this [ruby script in the inferno reference server](https://github.com/bulk-dqm/inferno-reference-server/blob/cms-122/upload_cms.rb)
    * After writing this script, I realized it’s pretty similar to the `uploadPremadeBundles` script. Open to feedback on whether it’s okay to keep this script or try to port over more logic from the existing script. For my given use case (uploading Synthea data), I would like to create groups for a given Synthea run rather than for a specific FHIR measure, and I would like to upload the practitioner/hospital batch bundles before uploading the transaction bundles for the rest of the patients. Also, I would like to use the updated ids that come with POSTing transaction bundles. This logic does not appear in the other script.

# Testing guidance
For all tests, use the following endpoint: `http://localhost:3001/`.
* POST a transaction bundle to the server and check for 200 OK response with a `transaction-response` resource
* POST a batch bundle to the server and check for 200 OK response with a `batch-response` resource
* POST a bundle with a different type (or POST a non-bundle resource) and check that an error is thrown
* POST a transaction bundle that throws an error (ex. One of the `request`s in the bundle is not of type POST or PUT) and check that an error is thrown
* Do the same for a batch bundle and check that no error is thrown, but that the problematic resource’s location is not included as an `entity` in the `batch-response` bundle that is returned. It should have a `response.outcome` instead
* If you want to test things end-to-end, create a directory of transaction bundles (ex. from synthea or elsewhere), run the script `npm run post-txn-bundles <bundle directory path> <group Id (will have “patients” appended to it after group creation)>`. Afterwards, take the group Id that is logged to the console and run a group `$export` on it. (So far tested with groups of up to 1000 patients)

### DIFF
--- a/src/scripts/postTransactionBundles.js
+++ b/src/scripts/postTransactionBundles.js
@@ -13,22 +13,26 @@ const mongoUtil = require('../util/mongo');
  */
 async function main() {
   const bundlePath = path.resolve(process.argv[2]);
+  const groupId = process.argv[3];
   const directoryFiles = fs.readdirSync(bundlePath);
   // store uploaded patientIds to be added as members to FHIR Group
   const patientRegEx = new RegExp('Patient/[^/]*');
 
   const patientIdsArray = directoryFiles.map(async file => {
-    const { data } = await axios.post(`http://localhost:3001/`,
-    JSON.parse(fs.readFileSync(path.join(bundlePath, file), 'utf8')), {headers: {'Content-Type': 'application/json+fhir'}});
-    
+    const { data } = await axios.post(
+      `http://${process.env.HOST}:${process.env.PORT}/`,
+      JSON.parse(fs.readFileSync(path.join(bundlePath, file), 'utf8')),
+      { headers: { 'Content-Type': 'application/json+fhir' } }
+    );
+
     const location = data.entry.find(e => e.response.location.startsWith('/Patient'));
     if (patientRegEx.test(location.response.location)) {
       return location.response.location.replace('/Patient/', '');
     }
   });
   const patientIds = await Promise.all(patientIdsArray);
-  createPatientGroupsPerMeasure('cms122-10', patientIds);
-  return `Group cms122-10-patients successfully created`;
+  createPatientGroupsPerMeasure(groupId, patientIds);
+  return `Group ${groupId}-patients successfully created`;
 }
 
 main()
--- a/src/scripts/postTransactionBundles.js
+++ b/src/scripts/postTransactionBundles.js
@@ -1,0 +1,37 @@
+const axios = require('axios');
+const path = require('path');
+const fs = require('fs');
+const { createPatientGroupsPerMeasure } = require('../util/groupUtils');
+const mongoUtil = require('../util/mongo');
+
+/**
+ * Uploads all transaction bundles from the specified directory
+ * into the database. Creates a group resource containing
+ * all the patients (with their updated ids that are created
+ * during the transaction bundle upload process) in the process.
+ * Uploads the created group resource to the database.
+ */
+async function main() {
+  const bundlePath = path.resolve(process.argv[2]);
+  const directoryFiles = fs.readdirSync(bundlePath);
+  // store uploaded patientIds to be added as members to FHIR Group
+  const patientRegEx = new RegExp('Patient/[^/]*');
+
+  const patientIdsArray = directoryFiles.map(async file => {
+    const { data } = await axios.post(`http://localhost:3001/`,
+    JSON.parse(fs.readFileSync(path.join(bundlePath, file), 'utf8')), {headers: {'Content-Type': 'application/json+fhir'}});
+    
+    const location = data.entry.find(e => e.response.location.startsWith('/Patient'));
+    if (patientRegEx.test(location.response.location)) {
+      return location.response.location.replace('/Patient/', '');
+    }
+  });
+  const patientIds = await Promise.all(patientIdsArray);
+  createPatientGroupsPerMeasure('cms122-10', patientIds);
+  return `Group cms122-10-patients successfully created`;
+}
+
+main()
+  .then(console.log)
+  .catch(console.error)
+  .finally(async () => await mongoUtil.client.close());
--- a/src/scripts/postTransactionBundles.js
+++ b/src/scripts/postTransactionBundles.js
@@ -5,7 +5,7 @@ const { createPatientGroupsPerMeasure } = require('../util/groupUtils');
 const mongoUtil = require('../util/mongo');
 
 // for each group, include the patients that are members of the previous group
-// (ex. creates a group of size 10, a group of size 100, ..., a group of size 20,000)
+// (ex. creates a group of size 10, a group of size 100, ..., a group of size 10,000)
 const GROUP_SIZES = [10, 90, 900, 9000];
 
 /**
--- a/src/scripts/postTransactionBundles.js
+++ b/src/scripts/postTransactionBundles.js
@@ -35,7 +35,7 @@ async function main() {
     // upload practitioner/hospital batch bundles first, if present
     directoryFiles
       .filter(file => file.startsWith('practitioner') || file.startsWith('hospital'))
-      .map(async file => {
+      .forEach(async file => {
         await axios.post(
           `http://${process.env.HOST}:${process.env.PORT}/`,
           JSON.parse(fs.readFileSync(path.join(directoryPath, file), 'utf8')),
--- a/src/scripts/postTransactionBundles.js
+++ b/src/scripts/postTransactionBundles.js
@@ -6,7 +6,7 @@ const mongoUtil = require('../util/mongo');
 
 // for each group, include the patients that are members of the previous group
 // (ex. creates a group of size 10, a group of size 100, ..., a group of size 20,000)
-const GROUP_SIZES = [10, 90, 900]; //, 9000, 1000];
+const GROUP_SIZES = [10, 90, 900, 9000];
 
 /**
  * Uploads all transaction bundles from the specified directory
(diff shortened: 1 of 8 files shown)


src/services/bundle.service.js
```diff
@@ -0,0 +1,149 @@
+const { v4: uuidv4 } = require('uuid');
+const { replaceReferences } = require('../util/bundleUtils');
+const { createResource, updateResource } = require('../util/mongo.controller');
+
+/**
+ * Creates transaction-response or batch-response bundle.
+ * Each entry elements SHALL contain a 'response' element which
```
elsaperelli, line 7:
```suggestion
 * Each entry element SHALL contain a 'response' element which
```
I think?

src/services/bundle.service.js
```diff
@@ -0,0 +1,149 @@
...
+    const { method } = entry.request;
+    return insertBundleResources(entry, method).catch(() => {
+      // exclude results for resource that produced an error, but
+      // continue executing remaining operations
+      return null;
+    });
+  });
+  const requestResults = await Promise.all(requestsArray);
+  return requestResults.filter(results => results !== null);
+};
```
elsaperelli, line 115: It really doesn't matter, but if you want to condense this a bit I think you can combine them and just have the stuff that is different in an if else statement? I only suggest this because there's a bit of overlap but it's all functional so doesn't matter :)

src/services/bundle.service.js
```diff
@@ -0,0 +1,149 @@
...
+ * Scrubs batch bundle entries and uploads each entry to the server.
+ * @param {Array} entries entries from POSTed batch bundle
+ * @returns array of request results
+ */
+const uploadResourcesFromBatchBundle = async entries => {
+  const scrubbedEntries = replaceReferences(entries);
+  const requestsArray = scrubbedEntries.map(async entry => {
+    const { method } = entry.request;
+    return insertBundleResources(entry, method).catch(() => {
+      // exclude results for resource that produced an error, but
```
elsaperelli, line 108: Out of curiosity, why is an error not thrown for the batch bundles? Is this something that's defined in the spec?

elsaperelli (changes requested):
Good stuff! I just have a couple suggestions/comments.

src/services/bundle.service.js
sarahmcdougall: Good point, I kept these separate because I was in a bit of a time crunch, but I agree we can definitely simplify this

src/services/bundle.service.js
sarahmcdougall: From the [spec](https://www.hl7.org/fhir/http.html#brules): "The success or failure of one change SHOULD not alter the success or failure or resulting content of another change...When processing the batch, the HTTP response code is 200 Ok if the batch was processed correctly, regardless of the success of the operations within the Batch. To determine the status of the operations, look inside the returned Bundle. A response code on an entry of other than 2xx (200, 202 etc) indicates that processing the request in the entry failed." 

I actually did not read that second part until now about including the entry in the bundle even if the response code is not 2xx... I'll look into adding that rather than skipping over the failed entries

elsaperelli (approved):
My only suggestion would be adding a link to the spec or a small blurb `bundle.service.js` about handling transaction v. batch bundle entries and why an error is thrown for one but not the other. Other than that, looks good to me! Nice job!

src/scripts/postTransactionBundles.js
```diff
@@ -0,0 +1,79 @@
...
+const GROUP_SIZES = [10, 90, 900, 9000];
+
+/**
+ * Uploads all transaction bundles from the specified directory
+ * into the database. Creates a group resource containing
+ * all the patients (with their updated ids that are created
+ * during the transaction bundle upload process) in the process.
+ * Uploads the created group resource to the database.
+ *
+ * Note: Synthea creates batch bundles for hospital/practitioner info, which
```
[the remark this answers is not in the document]
sarahmcdougall: Ah yeah, so Synthea creates batch bundles for hospital/practitioner info, and then there's an option to generate a Group resource containing the patient references (so not a bundle). Originally I had disabled that but the data Reece generated had the Group resource file included, so I added code to filter it out (since it has outdated references in it). I'll add a separate comment about it

src/scripts/postTransactionBundles.js
```diff
@@ -0,0 +1,81 @@
...
+  console.log('Connected successfully to server');
+
+  const bundleDir = path.resolve(process.argv[2]);
+
+  const patientRegEx = new RegExp('Patient/[^/]*');
+  // store uploaded patientIds to be added as members to FHIR Group (across all patients uploaded so far)
+  const allPatientIds = [];
+
+  for (const groupSize of GROUP_SIZES) {
+    const directoryPath = path.join(bundleDir, `cms122-${groupSize}-patients`, 'fhir');
```
natjoe4, line 32: Do we want to limit this to just cms122?

src/scripts/postTransactionBundles.js
```diff
@@ -0,0 +1,81 @@
...
+  await mongoUtil.client.connect();
+  console.log('Connected successfully to server');
+
+  const bundleDir = path.resolve(process.argv[2]);
+
+  const patientRegEx = new RegExp('Patient/[^/]*');
+  // store uploaded patientIds to be added as members to FHIR Group (across all patients uploaded so far)
+  const allPatientIds = [];
+
+  for (const groupSize of GROUP_SIZES) {
```
natjoe4, line 31: Will uploads from Synthea always have 10, 90, 900, and 9000 sized groups?

natjoe4 (changes requested):
Nice work. A few open questions and inline comments.

src/scripts/postTransactionBundles.js
sarahmcdougall: Generally they are not restricted to these sizes but for the script itself they will always be this size. I figure this isn't worth making an input argument right now, but we could make it one in the future if needed

src/util/groupUtils.js
```diff
@@ -0,0 +1,33 @@
+const { createResource } = require('./mongo.controller');
+
+/**
+ * Creates a FHIR Group resource that represents the Patients associated with a given Measure
+ * and adds it to the database
+ * @param {String} measureId An id for the associated FHIR Measure to be used in the Group Id
```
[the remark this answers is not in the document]
sarahmcdougall: Good catch, brought this over from the other file but agree we should change this

src/services/bundle.service.js
```diff
@@ -0,0 +1,131 @@
...
+    entry.resource.id = uuidv4();
+    const { id } = await createResource(entry.resource, entry.resource.resourceType);
+    if (id != null) {
+      entry.status = 201;
+      entry.statusText = 'Created';
+    }
+  }
+  if (method === 'PUT') {
+    const { id, created } = await updateResource(entry.resource.id, entry.resource, entry.resource.resourceType);
+    if (created === true) {
```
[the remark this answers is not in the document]
elsaperelli: I agree, but also the `=== true` might be good for readability so I have no preference.

elsaperelli (approved):
My only suggestion which is certainly not a huge priority but might be helpful is to add just a couple small unit tests? Maybe that's unnecessary though, up to you! Otherwise, looks great!

src/scripts/postTransactionBundles.js
```diff
@@ -0,0 +1,81 @@
...
+    const directoryPath = path.join(bundleDir, `cms122-${groupSize}-patients`, 'fhir');
+    const directoryFiles = fs.readdirSync(directoryPath);
+
+    // upload practitioner/hospital batch bundles first, if present
+    directoryFiles
```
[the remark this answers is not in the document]
natjoe4: I think I resolved this by accident. Still seems relevant

src/services/bundle.service.js
```diff
@@ -0,0 +1,131 @@
...
+  const requestsArray = scrubbedEntries.map(async entry => {
+    const { method } = entry.request;
+    return insertBundleResources(type, entry, method).catch(e => {
+      if (type === 'transaction') {
+        reply.code(400).send(e.message);
+      }
+    });
+  });
+  const requestResults = await Promise.all(requestsArray);
+  return requestResults;
```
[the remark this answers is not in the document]
natjoe4: Just bumping this in case it was forgotten :)

natjoe4 (changes requested):
Just a couple small reminders. Also, I agree with Elsa that unit tests would be ideal here, but those can be added in another task :)

PR #31, opened 2023-08-25
Add support for _typeFilter using search parameters

sarahmcdougall:
# Summary
Adds additional `_typeFilter` query support, now supporting a wide array of FHIR REST queries for each resource type using the projecttacoma fork of node-fhir-server-core and [Asymmetrik’s FHIR Query Builder](https://www.npmjs.com/package/@asymmetrik/fhir-qb).

## New behavior
The Bulk Export _typeFilter parameter, while still experimental, has the potential to greatly impact export performance. The _typeFilter parameter allows us to retrieve the tightest data set possible, as we can filter resource types by date fields, status, etc.

Right now, our server only supports code:in/type:in queries, and our _typeFilter query support is very specific to these kinds of queries. If we try to do another query, such as `MedicationRequest?status=active`, our server will just ignore this filter.

The updates in this PR allow the client to use _typeFilter values of the following formats:
•	Filter by search parameter for a given resource type (ex. `_typeFilter=MedicationRequest?status=active`)
•	Filter by multiple search parameters for a given resource type (ex. `_typeFilter=MedicationRequest?status=active&date=gt2018-07-01T00:00:00Z`)
•	Filter by conditional “or” search parameter queries for a given resource type (ex.  `_typeFilter=MedicationRequest?status=active,MedicationRequest?status=completed` to check for `MedicationRequest` resources with status “active” OR status “completed”)
•	Filter on multiple resource types via search parameters (ex. `_typeFilter=MedicationRequest? date=gt2018-07-01T00:00:00Z,Encounter?status=completed`

These queries can be used for system-level, group-level, and patient-level export.
## Code changes
**README**
- added a section on supported query parameters for the server

**Export.service.js**
- Wrapped all errors in `OperationOutcome` resources. The spec mentions in several places that the server should return an OperationOutcome resource in an error response when errors arise (ex. When an unsupported resource type is provided)
- Added a check on resource types that are included in the `_typeFilter` parameter

**ExportToNDJson.js**
Update `exportToNDJson` function to now do the following:
- Now acknowledges “search parameter” queries. These are `_typeFilter` queries that contain valid search parameters for the given resource type – for example, checking the `status` on `MedicationRequest` resources. These queries are indexed to grab the “property” and “property value” and pass them into the Asymmetrik FHIR Query Builder. For example, if the filter provided is `_typeFilter=MedicationRequest?status=completed&date=gt2018-07-01T00:00:00Z`, the function extracts the resource type (`MedicationRequest`) and extracts all subqueries (`status: “completed”` and `date`:  “gt2018-07-01T00:00:00Z”`).
- Continues to acknowledge the ValueSet-related `_typeFilter` queries that we currently support. An example query would be `_typeFilter=Encounter?type:in=http://some-value-set-url`. These queries are similarly parsed but are not passed through the query builder – instead, they get processed through the server to extract the codes from the ValueSet using existing server functions.

Updates `getDocuments` function to now do the following:
- For search parameter queries, runs an aggregation query to extract all the resources that match the queries. If ValueSet queries or patient filters also need to be applied, they are integrated into the aggregation pipeline.
- Miscellaneous changes to update `let` to `const` for many variables whose values never change, update variable names and function names to be more representative of their functionality, etc.

**valueSetHelper.js**
- Updates `getHierarchicalCodes` function to loop over an array of FHIR `ValueSet`s, as an array is passed into the function but is not appropriately accounted for within the function. This is a bug that was detected when testing the current functionality of the `ValueSet` filtering with `_typeFilter`

**Updated tests**
- export.server.test.js` tests now check for `OperationOutcome` responses
- `exportToNDJSON.test.js` changes
: (1) Create new Condition test fixture that contains two fields that can be filtered on , (2) Create new tests for `getDocuments` to check that it can filter based on multiple queries

# Testing guidance
First, run unit tests.
We will want to test a lot of combinations of `_typeFilter`, `_type`, and levels of export (system, group, and patient). Here’s an outline of steps that can be followed for testing:

1.	Upload resources to the server. This can be done with the `upload-bundles` script or `curl`. Make sure to upload groups of patients (the `upload-bundles` script does this automatically). The `post-txn-bundles` script can also be used to upload transaction bundles to the server, although it may need some tweaking to accommodate the group sizes you are uploading and your working directory.
2.	Run the server with `npm start`.
3.	Use Insomnia or a client like [the bulk data CLI Export Client](https://github.com/bulk-dqm/bulk-data-export-client) for sending requests to the server.

Requests to send to the server include, but are not limited to:
•	System-level export to ensure there are no errors
•	System-level export with `_type`
•	System-level export with `_typeFilter`
•	System-level export with both `_type` and `_typeFilter`

•	Group-level export to ensure there are no errors
•	Group-level export with `_type`
•	Group-level export with `_typeFilter`
•	Group-level export with both `_type` and `_typeFilter`

•	Patient-level export to ensure there are no errors
•	Patient-level export with `_type`
•	Patient -level export with `_typeFilter`
•	Patient -level export with both `_type` and `_typeFilter`

For the `_typeFilter` queries, test the following:
•	Query with “&” operator and check that the resources returned abide by all the conditions
•	Query with comma-delimited subqueries and check that the resources returned abide by at least one of the conditions
•	Test that the ValueSet filtering still works. To do this, you can refer to the data requirements of a given measure and extract a ValueSet url to include in the `_typeFilter` parameter. For example, you could do `_typeFilter=Encounter?type:in=<valueset-url>`

**Note**: Depending on the test environment that you are using, you may need to escape certain characters (like the `?` or `&` in the `_typeFilter` query) or manually URL encode them. Here are some cases I have come across when testing:
* From the terminal - you may need to escape `?` and `&` symbols (ex. `_typeFilter=MedicationRequest\?status=active`)
* From Insomnia - if you put the `_typeFilter` parameter and parameter value under the “Query” tab, you do not need to do any URL encoding - it gets picked up by Insomnia automatically. However, if you place the query within the request URL field (ex. Place `<url>/$export?_typeFilter=MedicationRequest?status=active` directly in the URL field), you will need to apply the URL encoding manually to the symbols, or else they will be interpreted to be separate parameters by Insomnia. (Ex. You may need to replace `&` with its URL encoding of `%26` in your request URL)

### DIFF
--- a/src/util/exportToNDJson.js
+++ b/src/util/exportToNDJson.js
@@ -13,21 +13,23 @@ const {
 } = require('./mongo.controller');
 const patientRefs = require('../compartment-definition/patient-references');
 const QueryBuilder = require('@asymmetrik/fhir-qb');
-const { getSearchParameters, resolveSchema } = require('@projecttacoma/node-fhir-server-core');
+const { getSearchParameters } = require('@projecttacoma/node-fhir-server-core');
 
 const qb = new QueryBuilder({ implementationParameters: { archivedParamPath: '_isArchived' } });
 
 /**
- * Uses Asymmetrik's getSearchParameters to retrieve the valid search parameters for a given
- * resource type. Builds a mapping of the search parameter x path to the parameter definition, which
- * contains the version, name, type, fhirtype, xpatah, definition, and description.
+ * Uses Asymmetrik's getSearchParameters function to retrieve the valid search parameters for a given
+ * resource type. Builds a mapping of the search parameter xpath to the parameter definition, which
+ * contains the version, name, type, fhirtype, xpath, definition, and description.
  * @param {string} resourceType FHIR resource type provided for export
  * @returns Record of valid search parameters and their associated data
  */
 const buildSearchParamList = resourceType => {
   const searchParams = {};
+  // get search parameters for FHIR Version 4.0.1
   const searchParameterList = getSearchParameters(resourceType, '4_0_1');
   searchParameterList.forEach(paramDef => {
+    // map xpath to parameter description
     {
       searchParams[paramDef.xpath.substring(paramDef.xpath.indexOf('.') + 1)] = paramDef;
     }
@@ -36,8 +38,9 @@ const buildSearchParamList = resourceType => {
 };
 
 /**
- * Exports the list of resources included in the _type member of the request object to NDJson
- * if the _type member doesn't exist it will simply export everything included in the supportedResources list
+ * Exports the list of resources included in the _type parameter to NDJSON, filtered according to the
+ * FHIR queries included in the _typeFilter parameter.
+ * If the _type parameter doesn't exist, the function will simply export all resource types included in the supportedResources list.
  * @param {string} clientId  an id to add to the file name so the client making the request can be tracked
  * @param {Array} types Array of types to be queried for, retrieved from request params
  * @param {string} typeFilter String of comma separated FHIR REST search queries
@@ -46,27 +49,31 @@ const buildSearchParamList = resourceType => {
  */
 const exportToNDJson = async (clientId, types, typeFilter, systemLevelExport, patientIds) => {
   try {
-    let dirpath = './tmp/';
+    const dirpath = './tmp/';
     fs.mkdirSync(dirpath, { recursive: true });
     let requestTypes = [];
     if (types) {
-      requestTypes = types; // this is the list types to export
+      // filter requested resource types to those specified with the _type parameter
+      requestTypes = types;
     } else {
       // create list of requested types if request.query._type param doesn't exist
       requestTypes.push(...supportedResources);
     }
-    let typefilterLookup = {};
+    // create lookup objects for (1) _typeFilter queries that contain search parameters, and (2) _typeFilter
+    // queries that contain type:in/code:in/etc. queries
+    const searchParameterQueries = {};
     const valueSetQueries = {};
     if (typeFilter) {
-      // Subqueries may be joined together with a comma for a logical "or"
-      let tyq = typeFilter.split(',');
-      // loop over each subquery
-      tyq.forEach(line => {
-        const resourceType = line.substring(0, line.indexOf('?'));
+      // subqueries may be joined together with a comma for a logical "or"
+      const tyq = typeFilter.split(',');
+      // loop over each subquery and extract all search params, which are joined via the "&" operator
+      // each subquery is of the format <resource type>?<query>
+      tyq.forEach(query => {
+        const resourceType = query.substring(0, query.indexOf('?'));
         // build mapping of search parameters for the given resource type
         const searchParams = buildSearchParamList(resourceType);
         // extract all properties that may be part of the same subquery via the "&" operator
-        const properties = line.substring(line.indexOf('?') + 1).split('&');
+        const properties = query.substring(query.indexOf('?') + 1).split('&');
         // create mapping of properties to their values within the given subquery
         const subqueries = {};
         properties.forEach(p => {
@@ -87,26 +94,31 @@ const exportToNDJson = async (clientId, types, typeFilter, systemLevelExport, pa
             subqueries[property] = propertyValue;
           }
         });
+        // pass all search parameter-related _typeFilter queries into the Asymmetrik query builder
         if (Object.keys(subqueries).length > 0) {
           const filter = qb.buildSearchQuery({
             req: { method: 'GET', query: subqueries, params: {} },
             parameterDefinitions: searchParams,
             includeArchived: true
           });
-          // TODO: add error handling here if a query cannot be built
           if (filter.query) {
-            if (typefilterLookup[resourceType]) {
-              typefilterLookup[resourceType].push(filter.query);
+            if (searchParameterQueries[resourceType]) {
+              searchParameterQueries[resourceType].push(filter.query);
             } else {
-              typefilterLookup[resourceType] = [filter.query];
+              searchParameterQueries[resourceType] = [filter.query];
             }
           }
         }
       });
     }
-    let docs = systemLevelExport ? requestTypes.filter(t => t !== 'ValueSet') : requestTypes;
-    docs = docs.map(async element => {
-      return getDocuments(element, typefilterLookup, valueSetQueries, patientIds);
+    const exportTypes = systemLevelExport ? requestTypes.filter(t => t !== 'ValueSet') : requestTypes;
+    let docs = exportTypes.map(async collectionName => {
+      return getDocuments(
+        collectionName,
+        searchParameterQueries[collectionName],
+        valueSetQueries[collectionName],
+        patientIds
+      );
     });
     docs = await Promise.all(docs);
     docs.forEach(doc => {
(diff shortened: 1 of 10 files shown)


elsaperelli (changes requested):
I did a bunch of testing and it all looks good to me so far! Just some small comments so far- going to continue testing a bit just to be sure.

src/util/exportToNDJson.js
```diff
@@ -8,13 +8,39 @@ const {
...
+      searchParams[paramDef.xpath.substring(paramDef.xpath.indexOf('.') + 1)] = paramDef;
+    }
+  });
+  return searchParams;
+};
+
+/**
+ * Exports the list of resources included in the _type parameter to NDJSON, filtered according to the
+ * FHIR queries included in the _typeFilter parameter.
+ * If the _type parameter doesn't exist, the function will simply export all resource types included in the supportedResources list.
```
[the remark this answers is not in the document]
sarahmcdougall: Good point - I added a sentence to the docstring to cover that

src/util/exportToNDJson.js
```diff
@@ -135,36 +225,35 @@ const writeToFile = function (doc, type, clientId) {
...
-    for (const propertyValue in typefilterLookupEntry) {
-      let results = typefilterLookupEntry[propertyValue].map(async value => {
+  if (valueSetQueries) {
+    for (const property in valueSetQueries) {
+      let results = valueSetQueries[property].map(async value => {
         let vs = await findOneResourceWithQuery({ url: value }, 'ValueSet');
+        // throw an error if we don't have the value set
         if (!vs) {
           throw new Error('Value set was not found in the database');
         }
```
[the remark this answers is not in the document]
sarahmcdougall: This error should be caught and bubbled up to an `OperationOutcome`, so we do not need to wrap it in an `OperationOutcome` object here. The function does not take in the `request`/`reply` objects, which would also make it difficult to throw the Operation Outcome directly from this function.

Here's the `OperationOutcome` I get back when I check for the bulk status:
Screenshot 2023-08-29 at 10 59 55 AM

test/services/export.service.test.js
```diff
@@ -210,7 +218,9 @@ describe('Check patient-level export logic (failure)', () => {
       .get('/Patient/$export?_unrecognizedparam=invalid')
       .expect(400)
       .then(response => {
-        expect(JSON.parse(response.text).message).toEqual(
+        expect(response.body.resourceType).toEqual('OperationOutcome');
+        expect(response.body.issue[0].code).toEqual(400);
+        expect(response.body.issue[0].details.text).toEqual(
```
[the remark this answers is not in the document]
sarahmcdougall: I added more tests to cover group-level failure and `_typeFilter` validation 👍

src/util/exportToNDJson.js
```diff
@@ -135,36 +225,35 @@ const writeToFile = function (doc, type, clientId) {
...
     stream.end();
   } else return;
 };
 /**
- * Processes the  entry in the _typeFilter and performs the lookup to determine if the type and code are present
- * in the listed value set.
- * @param {Object} typefilterLookupEntry an entry in the _typefilter to perform the valueSet lookup with
+ * Processes the entry in the _typeFilter parameter and performs the lookup to determine if the type and code are present
+ * in the listed ValueSet. Only relevant for _typeFilter queries that reference a ValueSet.
+ * @param {Object} valueSetQueries an entry in the _typefilter to perform the valueSet lookup with
```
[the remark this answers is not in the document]
sarahmcdougall: I updated the description of what the input parameter represents - let me know what you think! I'm open to suggestions for making further updates. 

For reference, the input will look something like this:
```typescript
{
  'type:in': [<array of ValueSets>],
  'code:in': [<array of ValueSets>]
}
```
I agree that maybe `valueSetQueries` is not the best name for what this collection represents, but I am having a hard time coming up with a decently descriptive var name here

elsaperelli (approved):
Thanks for adding those additional tests! I believe that helped coverage. Lgtm!

src/util/exportToNDJson.js
```diff
@@ -88,37 +156,61 @@ const exportToNDJson = async (clientId, types, typeFilter, systemLevelExport, pa
...
+      let query = q;
+      // wherever we have a $match, we need to add another "and" operator containing the ValueSet and/or patient query
+      if (Object.keys(vsQuery).length > 0) {
+        query.filter(q => '$match' in q).forEach(q => (q['$match'] = { $and: [q['$match'], vsQuery] }));
+      }
+      if (Object.keys(patQuery).length > 0) {
+        query.filter(q => '$match' in q).forEach(q => (q['$match'] = { $and: [q['$match'], patQuery] }));
+      }
+      // grab the results from aggregation - has metadata about counts and data with resources in the first array position
+      const results = (await findResourcesWithAggregation(query, collectionName, { projection: { _id: 0 } }))[0];
```
hossenlopp, line 198: I think there might be something wonky going on here. I have a database populated with ~112 Encounters most have `status: 'finished'`, but some don't. When I do an export with `_typeFilter=Encounter?status=finished` I'm only getting 10 Encounters. I think mongo isn't returning a complete result or is paging the result on us.

hossenlopp (changes requested):
I think mongo may be blocking the `_typeFilter` functionality from returning all resources that fit the query. We'll take a look at the cthon!

src/util/exportToNDJson.js
elsaperelli: Ahh good catch Hoss! I also noticed that when you do this export, you only get 10 of the other resources, regardless of how many there actually are in the database. This doesn't happen with `_type` export, so maybe the solution can be found there.

elsaperelli: It looks like `fhirQueryBuilder` has a `resultsPerPage` param whose default value is 10. We could change that to be something larger, but what should that value be?

sarahmcdougall: I will need to look into this more, but I think we could keep the default value to 10 and iteratively write the results to the file (iterating over each page)? Since we want to capture all possible results

elsaperelli: With some help with Hoss, I think we can iterate over each page with the following:

image

I have to do some more testing, but I believe this works!

sarahmcdougall: Tried this and the function ended up only going through the `while` loop once and still returning 10 results. I took a closer look at the actual query being used here, and `$facet.data` has a limit of 10, which I believe is the reason for still returning 10 results despite iterating over the cursor:
Screenshot 2023-09-13 at 11 24 44 AM

sarahmcdougall: Screenshot 2023-09-13 at 12 24 35 PM

Above is the `applyPaging` function in fhir-qb-mongo that gets used to build the query that is used for aggregation. To me, it seems like the logic for `numberOfPages` is sound, but the `$limit` may be restricting all the documents returned, which is causing only one page to exist? I will need to do some more investigating

elsaperelli: Ahhh I see! Sorry for the incorrect suggestion. I can also look into this 👍

sarahmcdougall: I don't think the suggestion is necessarily incorrect - I think it's good if we iterate over all the pages 👍 might just need some more investigation into `$limit` within the aggregation pipeline

sarahmcdougall:
Pushed updates to address the pagination support that is implemented by Asymmetrik.
For testing, I suggest doing the following:
1. Run an export request using `_type` on a given resource type to see how many resources exist of that resource type.
2. Run an export request using `_type` and `_typeFilter` on the same resource type. Try to use a `_typeFilter` query that will return > 10 resources, but < the total number of resources of the resource type.
3. Check that the amount of resources returned is greater than 10!

elsaperelli (approved):
Lgtm! Tested to make sure the new `$facet` workaround works

hossenlopp (approved):
This is an awesome first implementation. Smart to reuse work we have done elsewhere that used asymmetrik. I'm looking forward to building on this.

PR #32, opened 2023-12-06
Add support for "patient" parameter

sarahmcdougall:
# Summary
Adds support for the [patient parameter](https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters) for bulk data export.

## New behavior

The user can now send POST requests for all three types of export. The [bulk data kick-off request docs](https://build.fhir.org/ig/HL7/bulk-data/export.html#bulk-data-kick-off-request) note that “A server SHALL support GET requests and MAY support POST requests that supply parameters using the FHIR Parameters Resource.” The new `patient` parameter needs to be included via POST.

The user can now send a request with the `patient` parameter. The `patient` parameter is only applicable to patient-level and group-level export requests. When provided, the server SHALL NOT return resources in the patient compartment definition belonging to patients outside the list. The `patient` parameter is of type FHIR Reference, and multiple patients should be supported by the server. The `patient` parameter essentially acts as a patient-based filter in a similar way that group-level export applies patient-based filtering.

When the `patient` parameter is provided, the exported NDJSON is filtered based on the provided patient references. Resources that are in the patient compartment definition that do not reference the patients are not returned from the bulk export.

## Code changes
* Add support for POST requests - changes appear in `src/server/app.js`, `src/services/export.service.js` and in `test/services/export.service.test.js`
* `patient` is added the export queue jobs (`src/server/exportWorker.js`, `src/services/export.service.js`)
* Add functionality for retrieving parameters from a POST body (`gatherParams`) and validate the format of the `patient` param (`src/services/export.service.js`)
* Pass the `patient` param, if present, into the filtering logic (`src/util/exportToNDJson.js`)
        * Fortunately, the filtering logic in `exportToNDJson` already accounts for filtering via patient id, as it does so when ids are provided from a group. This logic is reused with the `patient` references, building an `$or` query for the references that are provided in the request.
* Various cases of error handling for edge cases (`patient` provided in a GET, patient does not exist on server, patient does not exist in group, etc.)
* Made updates to `ci.yml`: updates node version and mongo version, updates jest command for coverage report (based off the coverage report used for fqm-execution ci)

# Testing guidance
* Run unit tests
* An Insomnia import is attached to the pull request that contains requests that can be used for testing. These tests assume that the repository’s `upload-bundles` script has been run and that the server is loaded with groups created from that script. The tests are performed on the Cervical Cancer Screening measure, but the request urls and bodies can be tailored to other measures and parameter values.

A big concern with bulk data is the interaction between different parameters, and that should be kept in mind throughout testing. Make sure to test different combinations of `_type`, `_typeFilter`, and `patient`.
[bulk_patient_param_insomnia.json](https://github.com/projecttacoma/bulk-export-server/files/13590832/bulk_patient_param_insomnia.json)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ FHIR Operation to obtain a detailed set of FHIR resources of diverse resource ty
 
 Endpoint: `GET [fhir base]/Patient/$export`
 
-Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
+Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. The export parameters must be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
 
 #### All Patients in a Group
 
@@ -114,7 +114,7 @@ FHIR Operation to obtain a detailed set of FHIR resources of diverse resource ty
 
 Endpoint: `GET [fhir base]/Group/[id]/$export`
 
-Alternatively, a POST request (`POST [fhir base]/Group/[id]/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
+Alternatively, a POST request (`POST [fhir base]/Group/[id]/$export`) can be sent. The export parameters must be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
 
 #### System Level Export
 
@@ -122,7 +122,7 @@ Export data from a FHIR server, whether or not it is associated with a patient. 
 
 Endpoint: `GET [fhir base]/$export`
 
-Alternatively, a POST request (`POST [fhir base]/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
+Alternatively, a POST request (`POST [fhir base]/$export`) can be sent. The export parameters must be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
 
 For more information on the export endpoints, read this documentation on the [Export Request Flow](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow).
 
--- a/README.md
+++ b/README.md
@@ -106,17 +106,23 @@ FHIR Operation to obtain a detailed set of FHIR resources of diverse resource ty
 
 Endpoint: `GET [fhir base]/Patient/$export`
 
+Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
+
 #### All Patients in a Group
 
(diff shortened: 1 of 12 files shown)


src/services/export.service.js
```diff
@@ -208,17 +254,45 @@ function validateExportParams(request, reply) {
   return true;
 }
 
+/**
+ * Pulls query parameters from both the url query and request body and creates a new parameters map
```
elsaperelli, line 258: I have a question about the POST requests for group-level and patient-level exports. Correct me if I am wrong, but I believe the current functionality allows POST requests for group-level and patient-level exports with query params from either the request body or the url query. In the spec, it says "A server SHALL support GET requests and MAY support POST requests that supply parameters using the FHIR Parameters Resource" (https://www.hl7.org/fhir/parameters.html)." (https://build.fhir.org/ig/HL7/bulk-data/export.html#bulk-data-kick-off-request) Does this mean that POST requests should only be supported with query params in the request body as a FHIR Parameters resource? If so, we may want to add that check and throw an error like we do for the patient param in GET requests.

sarahmcdougall: Good catch - yes, I think we should assume that POST requests required a FHIR Parameters Resource as the request body for parameters if any params are specified. I will look into adding checks for the parameters being supplied in the body rather than in the url

sarahmcdougall: I added a check for parameters being supplied in the url, and a check for the request body to be of type "Parameters."  Open to adding additional handling here as well if any other edge cases come up while testing!

README.md
```diff
@@ -106,18 +106,24 @@ FHIR Operation to obtain a detailed set of FHIR resources of diverse resource ty
 
 Endpoint: `GET [fhir base]/Patient/$export`
 
+Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
```
elsaperelli, line 109:
```suggestion
Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. The export parameters must be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
```
Maybe change this since if it isn't supplied using a parameters resource then an error is thrown.

README.md
```diff
@@ -106,18 +106,24 @@ FHIR Operation to obtain a detailed set of FHIR resources of diverse resource ty
...
 
+Alternatively, a POST request (`POST [fhir base]/Group/[id]/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
+
 #### System Level Export
 
 Export data from a FHIR server, whether or not it is associated with a patient. This supports use cases like backing up a server, or exporting terminology data by restricting the resources returned using the `_type` parameter.
 
 Endpoint: `GET [fhir base]/$export`
 
+Alternatively, a POST request (`POST [fhir base]/$export`) can be sent. The export parameters should be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
```
elsaperelli, line 125:
```suggestion
Alternatively, a POST request (`POST [fhir base]/$export`) can be sent. The export parameters must be supplied using a FHIR [Parameters Resource](http://hl7.org/fhir/R4/parameters.html) in the request body.
```

elsaperelli (changes requested):
Tested again and everything looks good! Just one small wording change and then should be good to go.

elsaperelli (approved):
Lgtm!

PR #39, opened 2024-03-27
Add support for the _elements param

elsaperelli:
# Summary
This PR adds support for the [_elements](https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters) query parameter as defined by the Bulk Data Access IG. Some inspiration for this parameter's implementation was taken from the implementation of the `_elements` FHIR search parameter implementation in the measure-respository-service in [this PR](https://github.com/projecttacoma/measure-repository/pull/81).

## New behavior
Functionality for the `_elements` query parameter is now available! When the `_elements` query parameter is used, the server will return resources with only the elements specified in the parameter and the SUBSETTED tag (see the Bulk Data Access IG for more info). 

## Code changes
- `exportWorker.js` - add `elements` as a parameter to call to `exportToNDJson`
- `export.service.js` - add `_elements` parameter validation to `validateExportParams` function and add `_elements` as a recognized parameter
- `exportToNDJson.js` - add a try/catch to the `buildSearchParamList` function so that it doesn't error out on some resources that the server supports but for some reason do not exist in the node-fhir-server-core parameters (ex. `BiologicallyDerivedProduct`), add `_elements` parameter handling to `exportToNDJson` and `getDocuments`
- `mongo.controller.js` - add projection functionality to `findResourcesWithAggregation` so that we can use the `_elements` parameter with the `_typeFilter` parameter
- `exportToNDJson.test.js` - adds two tests- one for `_elements` handling in `getDocuments` and the other to make sure that `buildSearchParamList` doesn't error out on one of the `supportResources`

# Implementation Questions
- The Bulk Data Access IG defines the `_elements` parameter to include the following SHOULD functionality: "A server is not obliged to return just the requested elements. A server SHOULD always return mandatory elements whether they are requested or not." I do not have this implemented currently. I feel like it is something I should implement; however, it is an interesting question of how useful that may be (and potentially something we suggest in the IG?). I think it's interesting because in my opinion, if we defined mandatory elements to be elements that have a cardinality of 1..1, then there are quite a few and it may defeat the purpose of a potential use case of the `_elements` parameter (i.e., if a user just wanted the ids of all of the Condition resources).
- If JUST the element name is included, are we returning ALL resource types (and only include x where the element name x is included) OR are we returning ONLY resource types where the element name x is included (right now we do the former). 

# Testing guidance
- Please read and consider my implementation questions! 
- `npm run check` 
- `npm start` to spin up the server at `localhost:3001`
- Try lots of different HTTP request combinations! Try all three endpoints (System-level, Patient, Group) and combinations of the `_elements` parameter with other query parameters (`_type`, `patient`, `_typeFilter`, etc.).
- I would try making your own HTTP requests in Insomnia first because it may bring up something that I missed, but I also have the following Insomnia tests attached!

[ElementsParamTesting.json](https://github.com/projecttacoma/bulk-export-server/files/14780950/ElementsParamTesting.json)

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ The server supports the following query parameters:
 
 #### `_elements` Query Parameter
 
-The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type have a minimum cardinality of 1. Because this server provides json-formatted data, `resourceType` is also an implied mandatory element for all Resources.
+The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements used to filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type which have a minimum cardinality of 1. Because this server provides json-formatted data, `resourceType` is also an implied mandatory element for all Resources.
 
 The returned resources are only filtered by elements that are applicable to them. For example, if a request looks like the following:
 
--- a/README.md
+++ b/README.md
@@ -154,11 +154,11 @@ The server supports the following query parameters:
 - `_outputFormat`: The server supports the following formats: `application/fhir+ndjson`, `application/ndjson+fhir`, `application/ndjson`, `ndjson`
 - `_typeFilter`: Filters the response to only include resources that meet the criteria of the specified comma-delimited FHIR REST queries. Returns an error for queries specified by the client that are unsupported by the server. Supports queries on the ValueSets (`type:in`, `code:in`, etc.) of a given resource type.
 - `patient`: Only applicable to POST requests for group-level and patient-level requests. When provided, the server SHALL NOT return resources in the patient compartment definition belonging to patients outside the list. Can support multiple patient references in a single request.
-- `_elements`: Filters the content of the responses to omit unlisted, non-mandatory elements form the resources returned. These elements should be provided in the form `[resource type].[element name]` (e.g., `Patient.id`) which only filters the contents of those specified resources or in the form `[element name]` (e.g., `id`) which filters the contents of all of the returned resources.
+- `_elements`: Filters the content of the responses to omit unlisted, non-mandatory elements from the resources returned. These elements should be provided in the form `[resource type].[element name]` (e.g., `Patient.id`) which only filters the contents of those specified resources or in the form `[element name]` (e.g., `id`) which filters the contents of all of the returned resources.
 
 #### `_elements` Query Parameter
 
-The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type have a minimum cardinality of 1.
+The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type have a minimum cardinality of 1. Because this server provides json-formatted data, `resourceType` is also an implied mandatory element for all Resources.
 
 The returned resources are only filtered by elements that are applicable to them. For example, if a request looks like the following:
 
@@ -174,7 +174,7 @@ If a request does not specify a resource type, such as the following:
 GET http://localhost:3000/$export?_elements=id
 ```
 
-Then the returned resources should only contain an `id` (if applicable) and any mandatory elements.
+Then all returned resources should only contain an `id` (if applicable) and any mandatory elements.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -154,6 +154,27 @@ The server supports the following query parameters:
 - `_outputFormat`: The server supports the following formats: `application/fhir+ndjson`, `application/ndjson+fhir`, `application/ndjson`, `ndjson`
 - `_typeFilter`: Filters the response to only include resources that meet the criteria of the specified comma-delimited FHIR REST queries. Returns an error for queries specified by the client that are unsupported by the server. Supports queries on the ValueSets (`type:in`, `code:in`, etc.) of a given resource type.
 - `patient`: Only applicable to POST requests for group-level and patient-level requests. When provided, the server SHALL NOT return resources in the patient compartment definition belonging to patients outside the list. Can support multiple patient references in a single request.
+- `_elements`: Filters the content of the responses to omit unlisted, non-mandatory elements form the resources returned. These elements should be provided in the form `[resource type].[element name]` (e.g., `Patient.id`) which only filters the contents of those specified resources or in the form `[element name]` (e.g., `id`) which filters the contents of all of the returned resources.
+
+#### `_elements` Query Parameter
+
+The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type have a minimum cardinality of 1.
+
+The returned resources are only filtered by elements that are applicable to them. For example, if a request looks like the following:
+
+```
+GET http://localhost:3000/$export?_elements=Condition.id
+```
+
+Then the returned resources should contain everything on them except the returned Conditions should only contain an `id` (if applicable) and any mandatory elements.
+
+If a request does not specify a resource type, such as the following:
+
+```
+GET http://localhost:3000/$export?_elements=id
+```
+
+Then the returned resources should only contain an `id` (if applicable) and any mandatory elements.
 
 ## License
 
--- a/src/scripts/parseStructureDefinitions.js
+++ b/src/scripts/parseStructureDefinitions.js
@@ -23,11 +23,10 @@ async function main() {
 
     // read the contents of the file
     const structureDef = JSON.parse(fs.readFileSync(f.fullPath, 'utf8'));
-    // QUESTION: should I be using snapshot or differential ?
     structureDef.snapshot.element.forEach(e => {
       const elem = e.id.split('.');
       if (elem.length === 2) {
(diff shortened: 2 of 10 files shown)


lmd59 (commented):
Initial code changes make sense, though understood we're still waiting on implementation details. I think it makes sense to do a README update to the server endpoints section for this PR as well. (When you do, there are a couple of deqm-test-server references that might be nice to cleanup to reference bulk-export-server instead)

lmd59:
Implementation question 1: 
My initial thoughts are that it might be worth implementing the "mandatory elements". The mandatory elements could be helpful in an _elements param request if the requester is expecting back valid FHIR resources, which I could see being a requirement in certain use cases. And while I think the option to not include mandatory elements also has valid use cases, I think at least attempting to implement it would give us more information in our goal of "exercising the spec". If we do end up making an IG recommendation, this would give us more data to go off of, whether we recommend in favor or against mandatory elements.

Implementation question 2:
This is potentially dependent on 1. If we're just returning elements requested (and not mandatory ones), then it probably wouldn't make sense to return resources that don't have at least one of the elements requested. If we're also returning mandatory elements, I would lean toward returning all resource types and assuming the burden of filtering should be on the requester.

lmd59 (commented):
I have a couple of general comments and questions in code and then I'll also add:
1. It seems like `resourceType` is implicitly mandatory on the base resource element, but the structure definition itself doesn't indicate that - https://hl7.org/fhir/R4/resource.html . Do you think we would consider that mandatory for all resources?
2. I would add a description of the _elements param to the README

elsaperelli:
> I have a couple of general comments and questions in code and then I'll also add:
> 
> 1. It seems like `resourceType` is implicitly mandatory on the base resource element, but the structure definition itself doesn't indicate that - https://hl7.org/fhir/R4/resource.html . Do you think we would consider that mandatory for all resources?
> 2. I would add a description of the _elements param to the README

Good question about `resourceType`! Yeah I am confused on this as well- it looks like it is required in the TypeScript FHIR definition which is interesting, but as you said, does not appear in the structure definition? I am going to defer to @hossenlopp and @p9g for this

src/services/export.service.js
```diff
@@ -223,6 +232,55 @@ function validateExportParams(parameters, reply) {
...
 
+  // add validation for the _elements query param
+  if (parameters._elements) {
+    const elementsArray = parameters._elements.split(',');
+    const unsupportedResourceTypes = [];
+    const unsupportedElementTypes = [];
+    elementsArray.forEach(line => {
+      // split each of the elements up by a '.' if it has one. If it does, the first part is the resourceType and the second is the element name
+      // if there is no '.', we assume that the element is just the element name
+      // TODO: add some sort of check for unsupported elements
```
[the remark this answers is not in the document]
elsaperelli: I think I will take them out- after our discussion today I think we decided that we don't need to check for the support of element names (we don't for `_typeFilter`). We can always reassess this in the future.

src/scripts/parseStructureDefinitions.js
```diff
@@ -0,0 +1,48 @@
...
+  files.forEach(f => {
+    let mandatoryElements = [];
+
+    // read the contents of the file
+    const structureDef = JSON.parse(fs.readFileSync(f.fullPath, 'utf8'));
+    // QUESTION: should I be using snapshot or differential ?
+    structureDef.snapshot.element.forEach(e => {
+      const elem = e.id.split('.');
+      if (elem.length === 2) {
+        if (e.min === 1) {
```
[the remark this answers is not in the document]
elsaperelli: Doesn't hurt to add! Nice catch!

src/scripts/parseStructureDefinitions.js
```diff
@@ -0,0 +1,48 @@
+const fs = require('fs');
+const path = require('path');
+
+const STRUCTURE_DEFINITIONS_BASE_PATH = path.join(__dirname, '../resource-definitions');
```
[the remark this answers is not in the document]
elsaperelli: Good call!

p9g:
https://hl7.org/fhir/R4/json.html#resources says "A resource is a JSON object with a property resourceType which informs the parser which resource type this is"  .
So if it does not have resourceType, it is not a FHIR resource.  Sounds mandatory.  
http://hl7.org/fhir/R4/json.html#xml comparison of json and xml forms of a resource: "The type of the resource is represented differently in JSON - instead of being the name of the base object (there is none in JSON), it is carried as the property resourceType"

The json schema http://hl7.org/fhir/R4/fhir.schema.json.zip has resourceType as required for all of the (duh) resources.

elsaperelli:
Huge thank you for the guidance as well as Karl's! Looks like our current implementation makes the most sense although it certainly can be subject to change.

For testing now, please refer to Karl's guidance and make sure this functionality aligns with his suggestions. Also please check that my README explanation makes sense and if there should be anything else added.

I have an updated Insomnia testing collection attached! Please use this one now that `resourceType` is a mandatory element and the endpoint is `localhost:3000`. 

[ElementsParamTesting.json](https://github.com/projecttacoma/bulk-export-server/files/15095520/ElementsParamTesting.json)

README.md
```diff
@@ -154,6 +154,27 @@ The server supports the following query parameters:
...
 - `patient`: Only applicable to POST requests for group-level and patient-level requests. When provided, the server SHALL NOT return resources in the patient compartment definition belonging to patients outside the list. Can support multiple patient references in a single request.
+- `_elements`: Filters the content of the responses to omit unlisted, non-mandatory elements form the resources returned. These elements should be provided in the form `[resource type].[element name]` (e.g., `Patient.id`) which only filters the contents of those specified resources or in the form `[element name]` (e.g., `id`) which filters the contents of all of the returned resources.
+
+#### `_elements` Query Parameter
+
+The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type have a minimum cardinality of 1.
```
lmd59, line 161:
```suggestion
The server supports the optional and experimental query parameter `_elements` as defined by the Bulk Data Access IG (here)[https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters]. The `_elements` parameter is a string of comma-delimited HL7® FHIR® Elements used to filter the returned resources to only include listed elements and mandatory elements. Mandatory elements are defined as elements in the StructureDefinition of a resource type which have a minimum cardinality of 1. Because this server provides json-formatted data, `resourceType` is also an implied mandatory element for all Resources.
```

README.md
```diff
@@ -154,6 +154,27 @@ The server supports the following query parameters:
...
+
+Then the returned resources should contain everything on them except the returned Conditions should only contain an `id` (if applicable) and any mandatory elements.
+
+If a request does not specify a resource type, such as the following:
+
+```
+GET http://localhost:3000/$export?_elements=id
+```
+
+Then the returned resources should only contain an `id` (if applicable) and any mandatory elements.
```
lmd59, line 177:
```suggestion
Then all returned resources should only contain an `id` (if applicable) and any mandatory elements.
```

lmd59 (commented):
Just some recommendations for the README. Otherwise, looks good!

lmd59 (approved):

hossenlopp (commented):
This looks like it works well for the most part, but if you reference a choice type parameter such as `DiagnosticReport.effective` you do not get the FHIR JSON `effectiveDateTime` and `effectivePeriod` choice elements. The spec isn't clear if it should reference the choice types by the name of the element or the element and the specific type.

elsaperelli:
> This looks like it works well for the most part, but if you reference a choice type parameter such as `DiagnosticReport.effective` you do not get the FHIR JSON `effectiveDateTime` and `effectivePeriod` choice elements. The spec isn't clear if it should reference the choice types by the name of the element or the element and the specific type.

Good call! I reached out to Karl and confirmed that this is indeed the functionality that we want for our current interpretation of the spec. I added two files: `getChoiceTypesForResource.js` which is a script for parsing the FHIR R4 StructureDefinitions for each resource type supported by the server and getting the choice type elements and their types for each and `choice-types.json` which is a lookup object that is a result of running the `getChoiceTypesForResource.js` script. 

New behavior for the server now includes the following: for `_elements` parameter inputs, if the element name is a choice type for a resource, then the server will return any of the choice types. For example, `DiagnosticReport.effective` should return DiagnosticReport resources that contain `effectiveDateTime` and `effectivePeriod`. The `ecqm-content-r4-2021` content seems to not have `effectivePeriod` on the DiagnosticReport resources, so a better test may be MedicationRequest.medication should return medicationCodeableConcept. I will update the Insomnia collection to include GET requests for this cases.

src/util/exportToNDJson.js
```diff
@@ -114,6 +120,62 @@ const exportToNDJson = async (clientId, types, typeFilter, patient, systemLevelE
...
+        let elementNames = [];
+        if (e.includes('.')) {
+          const resourceType = e.split('.')[0];
+          const elementName = e.split('.')[1];
+          if (
+            Object.keys(choiceTypeElements[resourceType]).length !== 0 &&
+            Object.keys(choiceTypeElements[resourceType]).includes(`${elementName}[x]`)
+          ) {
+            const rootElem = elementName.split('[x]')[0];
+            choiceTypeElements[resourceType][rootElem].forEach(e => {
```
hossenlopp, line 137: This doesn't carefully check the choice type element structure exists before doing the `forEach`. 

It blows up on the following request
```
http://localhost:3000/$export?_type=DiagnosticReport&_elements=DiagnosticReport.effective
```

Additionally this area of the code could use a bit more explanation of what it is doing. If possible, unit tests too.

hossenlopp (changes requested):
More null checks on the choice type stuff is needed.

src/util/exportToNDJson.js
elsaperelli: Sounds good! I think I have a fix that I will push up. The issue was it wasn't finding the key with just the elementName in the lookup object (needed the [x] as well). I will add more documentation but will push that fix for now!

hossenlopp: I think it also needs to be a bit more resilient to garbage being passed in. Like a Resource or element that doesn't exist.

elsaperelli: How so? Should I check every resourceType and element that is passed in to make sure it's valid FHIR? If so, how does that work for elements passed in without resourceType (i.e. just `id` rather than `Conditiion.id`)? Would I have to check that it's valid on every single resourceType (often not the case)? These are things I have brought up but it seemed like the server should not error out.

lmd59: If you wanted to do a base check that something of the form `Condition.id` has id as a valid property on Condition, you could use the propertyPaths file/script here: https://github.com/projecttacoma/fqm-testify/commit/815c4fc5cef44f3f4056ce873f1b3678549fdb9b
Hasn't been pulled into fqm-testify because it's mostly just been used for the data requirements work so far.

lmd59: Do we have an idea of the expected error output according to the spec if there is malformed parameters?

elsaperelli: @hossenlopp @lmd59 also keep in mind that we don't do any sort of checks for inputs passed into `_typeFilter`. For example, a GET request to `$export?_typeFilter=Condition?invalid=invalid` does not return any errors.

lmd59: I think maybe some of this validation is out of scope for this tasking, but if we're not responding with the correct errors, we should know what the spec indicates we need and use that for potential follow-up tasking.

lmd59: Probably just 400 Bad Request, right?

hossenlopp (approved):
This works well!

lmd59 (approved):

PR #41, opened 2024-04-10
Add docker setup for bulk export

elsaperelli:
# Summary
This PR sets up a working Docker environment for this server and changes some things so that it can be run on our abacus-dev ECE.

## New behavior
Server can be run with Docker! Also pushed this repository to docker hub and added it to abacus-dev.

## Code changes
- `.env` - adds `PUBLIC_BULK_SERVER` variable and changes `PORT` to 3000 for consistency
- `.env.test` - adds `PUBLIC_BULK_SERVER` variable
- `Dockerfile` - upgrades node, adds `ENV HOST` and `ENV PORT` variables
- `docker-compose.example.yml` - example docker-compose for future reference (similar to that in the `measure-repository`)
- `docker-compose.yml` - add redis, upgrade mongo, add `PUBLIC_BULK_SERVER`
- `postTransactionBundles.js`/`bulkstatus.service.js`/`export.service.js`/`bulkstatus.service.test.js` - use `PUBLIC_BULK_SERVER`

# Testing guidance
- `npm run check`
- Start up the server with docker with `docker compose up --build`
- Use Insomnia to ensure functionality is the same as if the server was spun up locally. See PR #39 for example Insomnia requests to get started.
- Start the server locally to ensure that all runs the same.


lmd59 (commented):
Testing through things a bit more, but a couple of small initial suggestions

docker-compose.example.yml
```diff
@@ -0,0 +1,40 @@
...
+      DB_PORT: 27017
+      DB_NAME: bulk-export-server
+      REDIS_HOST: redis
+      REDIS_PORT: 6379
+      EXPORT_WORKERS: 2
+    ports:
+      - '3000:3000'
+    volumes:
+      - ./src:/usr/src/app/src
+    command: npm start
```
hossenlopp, line 22: As mentioned in `docker-compose.yml` this isn't need and the `command` should have been specified in the Dockerfile and also not needed in the compose.
```suggestion
    volumes:
      - ./src:/usr/src/app/src
    command: npm start
```

docker-compose.example.yml
```diff
@@ -0,0 +1,40 @@
...
+      - redis
+    environment:
+      # Change this for public location of bulk-export-server this should be the FQDN and location of where the
+      # bulk-export container is made public to users. ex. https://abacus.example.com/bulk-export
+      PUBLIC_BULK_SERVER: http://localhost:3000
+      DB_HOST: mongo
+      DB_PORT: 27017
+      DB_NAME: bulk-export-server
+      REDIS_HOST: redis
+      REDIS_PORT: 6379
```
hossenlopp, line 16: Suggest moving the environment variables for the default ports into the Dockerfile so they don't need to be here.

```suggestion
      DB_HOST: mongo
      REDIS_HOST: redis
```

docker-compose.yml
```diff
@@ -4,25 +4,38 @@ services:
...
     build:
       context: .
       dockerfile: Dockerfile
     environment:
+      PUBLIC_BULK_SERVER: http://localhost:3000
       DB_HOST: mongo
       DB_PORT: 27017
       DB_NAME: bulk-export-server
+      REDIS_HOST: redis
+      REDIS_PORT: 6379
```
hossenlopp, line 17: Suggest moving the default ports to the Dockerfile.
```suggestion
      DB_HOST: mongo
      DB_NAME: bulk-export-server
      REDIS_HOST: redis
```

hossenlopp (changes requested):
Some of the stuff in the docker-compose can be paired down and moved to the Dockerfile.

Suggest renaming `PUBLIC_BULK_SERVER` to something that implies it is the base URL. Perhaps `BULK_BASE_URL`.

We could also use a script for running and pushing the build like we have in `measure-repository` service.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16
```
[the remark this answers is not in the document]
elsaperelli: True true!

docker-compose.example.yml
elsaperelli: Do I still need the `DB_NAME` here?

hossenlopp: That's a good point. The default could be kept in the Dockerfile

elsaperelli:
@hossenlopp I first misread your comment thinking you asked for a bundle upload script...but now we have both! Added a `directory-upload.sh` script and a `docker-build.sh` script like we do in the measure-repository.

hossenlopp (approved):

Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM node:14
+FROM node:18 as base
+
+FROM base as deps
```
lmd59, line 3: Maybe it's a convention? But is there a reason that we immediately change to a different named build step without doing anything with base?
It seems like we could condense these first two lines to:
```suggestion
FROM node:18 as deps
```
because we're not using base anywhere else (we're using node:18-slim instead to run the app below).

lmd59 (commented):
One small comment/question. Otherwise I think it looks good!

lmd59 (approved):

lmd59 (approved):

hossenlopp (approved):

PR #45, opened 2024-06-18
Create Patient Service and Implement CRUD operations

cgolemme:
# Summary

This PR adds endpoints to the bulk-export-server to handle operations related to Patients. Adds operations to create, get, update and delete patients on the server. This PR also adds `Group` delete endpoint. `[host]/Patient` and `[host]/Group` now correctly return a `searchset` as per fhir spec, https://hl7.org/fhir/R4/http.html#search. 

[JIRA Issue](https://jira.mitre.org/browse/TACOSTRAT-1506)

## New behavior

bulk-export-server now has CRUD operations for Patients
- `GET` : `[host]/Patient`
- `GET` : `[host]/Patient/[patientId]`
-  `POST` : `[host]/Patient`
-  `PUT` : `[host]/Patient/[patientID]`
- `DELETE` : `[host]/Patient/[patientID]`

Now has new delete operation for Group
- `DELETE` : `[host]/Group/[groupID]`

Endpoints that find multiple Patients or Groups from the server, `GET` : `[host]/Patient` and `GET` : `[host]/Group`, now return a searchset object instead of a json list of fhir Patient or Group objects.

## Code changes
- `.eslintrc.json` and `.prettierignore` were changed to add `ecqm-content-r4-2021` to the excluded folders because they were causing warnings with eslint and prettier.
- `src/server/app.js` was changed to add the new patient endpoints and new group endpoint to the server.
- `src/services/group.service.js`: updated some comments to be more specific, wrapped the /Group get endpoint in a search set object, added new function for DELETE endpoint.
- `src/services/patient.service.js` was created to implement functions that will interact with the database to fulfill the CRUD operations requested when a user accesses the endpoints for Patients.
- `test/fixtures/updatedTestPatient.json` was created to test the `PUT` endpoint.
- `test/services/patient.service.test.js` are tests for the patient.service.js operations that follows closely from the `group.service.test.js` file.
- `src/util/bundleUtils.js`: created to generate the searchset bundle object that wraps the response from the server for the /Patient and /Group endpoints.
- `src/util/mongo.controller.js`: added a default projection to filter out mongo-generated ids from the server's responses to all operations.
- `test/services/group.service.test.js`: changed /Group test to interact with new searchset response from server

# Testing guidance
- run `npm run check`
- Check that prettier and eslint cause no errors
- Check that all the tests pass
- Test endpoints on Insomnia

### DIFF
--- a/src/util/mongo.controller.js
+++ b/src/util/mongo.controller.js
@@ -26,7 +26,7 @@ const createResource = async (data, resourceType) => {
  */
 const findResourceById = async (id, resourceType) => {
   const collection = db.collection(resourceType);
-  return collection.findOne({ id: id });
+  return collection.findOne({ id: id }, { projection: { _id: 0 } });
 };
 
 /**
@@ -37,10 +37,11 @@ const findResourceById = async (id, resourceType) => {
  */
 const findOneResourceWithQuery = async (query, resourceType) => {
   const collection = db.collection(resourceType);
-  return collection.findOne(query);
+  return collection.findOne(query, { projection: { _id: 0 } });
 };
 
-const findResourcesWithQuery = async (query, resourceType, options = {}) => {
+const findResourcesWithQuery = async (query, resourceType, options = { projection: { _id: 0 } }) => {
+  options.projection['_id'] = 0;
   const collection = db.collection(resourceType);
   const results = collection.find(query, options);
   return results.toArray();
--- a/src/util/mongo.controller.js
+++ b/src/util/mongo.controller.js
@@ -40,8 +40,13 @@ const findOneResourceWithQuery = async (query, resourceType) => {
   return collection.findOne(query, { projection: { _id: 0 } });
 };
 
+/**
+ * Searches the database for the one or more resources based on a mongo query and returns the data.
+ * @param {Object} query the mongo query to use
+ * @param {string} resourceType type of desired resource, signifies collection resource is stored in
+ * @return {Array} the data of the found documents
+ */
 const findResourcesWithQuery = async (query, resourceType, options = { projection: { _id: 0 } }) => {
-  options.projection['_id'] = 0;
   const collection = db.collection(resourceType);
   const results = collection.find(query, options);
   return results.toArray();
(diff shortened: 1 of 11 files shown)


elsaperelli (approved):
Lgtm!

The only thing I would change in the future (not a big deal for this one at all) but just make sure that you include in the testing guidance to try out the new endpoints in Insomnia! (This should be kinda obvious but good to add)

hossenlopp (changes requested):
The listing of all patients is returning a JSON Array this should be a `searchset` Bundle instead. See https://hl7.org/fhir/R4/http.html#search for more info.

Additionally we should have the internal mongo `_id`s projected out of the responses.

cgolemme:
I made the requested changes. One potentially weird thing that I did was call `const found = await findResourceById(request.params.groupId, 'Group');` in the group and patient delete functions in order to correctly return the 404 that we want because there is nowhere in the call stack of `removeReasource` that could throw an error the resource to remove is not found.

src/util/mongo.controller.js
```diff
@@ -37,10 +37,11 @@ const findResourceById = async (id, resourceType) => {
  */
 const findOneResourceWithQuery = async (query, resourceType) => {
   const collection = db.collection(resourceType);
-  return collection.findOne(query);
+  return collection.findOne(query, { projection: { _id: 0 } });
 };
 
-const findResourcesWithQuery = async (query, resourceType, options = {}) => {
+const findResourcesWithQuery = async (query, resourceType, options = { projection: { _id: 0 } }) => {
+  options.projection['_id'] = 0;
```
elsaperelli, line 44: I am not sure if we need to pass in the projection and then set it again? Honestly seems like we could probably even get rid of the `options = {}` input (I know that was already there but looks like it isn't even needed lol) and then you can just do `const results = collection.find(query, { projection: { _id: 0 } });` unless I am missing something (very possible).

elsaperelli (changes requested):
Just one small comment but otherwise almost good to go!

elsaperelli (approved):
Lgtm! Way to be flexible!!

hossenlopp (approved):

PR #47, opened 2024-06-28
Kickoff import

lmd59:
# Summary
Creates a new endpoint on the bulk-export-server to kickoff $import against a specified endpoint on an import server

## New behavior
Server now supports endpoint: `POST [fhir base]/bulkstatus/[client id]/kickoff-import`

## Code changes
- Update README to describe new endpoint
- Update package.json to make axios play nice with jest
- Add new endpoint to app.js and bulkstatus.service.js, including pulling data from the existing bulk status to create the import manifest, appropriate error handling, and appropriate pass-through of headers
- Pull `gatherParams` out of export.service.js into its own shared serviceUtils

# Testing guidance

- `npm run check`
- Start bulk-export-server on port 3001 (and change to port 3001 for both environment variables)
- Start deqm-test-server (on default port 3000, make sure you have the latest with the new $import workflow)
- The Insomnia collection below has a basic set of requests that go through the workflow as follows:
  - Do an $export request on bulk-export-server
  - Check the status of that request at the bulk-export-server bulkstatus endpoint (from the $export response content-location header)
  - Add `/kickoff-import` to the end of the bulkstatus request and use that endpoint to do a POST with a receiver parameter in the body. Example:
`{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "receiver",
			"valueString": "http://localhost:3000/4_0_1/$import"
    }
  ]
}`
  - Check the status of the $import request using the deqm-test-server bulkstatus endpoint (from the kickoff-import response content location header)
  - Repeat this workflow for other export options and for error options to make sure the kickoff-import responses makes sense

[kickoff-import-insomnia.json](https://github.com/user-attachments/files/16036594/kickoff-import-insomnia.json)

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,16 @@ Alternatively, a POST request (`POST [fhir base]/$export`) can be sent. The expo
 
 For more information on the export endpoints, read this documentation on the [Export Request Flow](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow).
 
+#### Bulk Status
+
+This server supports the bulk status endpoint in support of the [Export Request Flow](https://hl7.org/fhir/uv/bulkdata/export/index.html#request-flow).
+
+Endpoint: `GET [fhir base]/bulkstatus/[client id]`
+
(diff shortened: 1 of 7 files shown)


elsaperelli (changes requested):
Just one small comment! I think we can wait to think about unit testing for this when we do the unit testing task for the `bulk-export-server`.

elsaperelli (approved):
Lgtm!

PR #49, opened 2024-08-07
ValueSet export is Bad Request

lmd59:
# Summary
Previously a system export with `_type=ValueSet` had an unexpected file read error. ValueSet should not be allowed to be exported, so 

## New behavior
Any export request with ValueSet as the type specified in the parameters gives a Bad Request response

## Code changes

- export.service update to exclude ValueSet from the types that are supported for export (but not from the types that are supported for storage in the collection creation script)
- exportToNDJson change to properly create the full client id folder to prevent other issues like this

# Testing guidance

- `npm run check`
- `npm run start`
- Insomnia export requests such as `$export?_typeFilter=ValueSet?id=test`, `$export?_elements=ValueSet.id`, `$export?_type=ValueSet` should all return a 400 specifying ValueSet is unsupported.


elsaperelli (approved):
Lgtm!

PR #50, opened 2024-08-16
Flexporter mapping for 130 and initial README

lmd59:
# Summary
Adds initial flexporter mapping for qicore which addresses specific data elements required for CMS130. Also adds README explaining initial flexporter mapping process and recommendations for the future.
Note: Recommended future tasking could involve creating a script to pull synthea in here and run flexporter mappings for bulk data generation. Some scripting could also be created to streamline the mapping creation process.

## New behavior
Synthea run with flexporter mapping creates data that is conformant enough to qicore for the purposes of CMS130 calculation.

## Code changes
Creates flexporter mapping files and README file.

# Testing guidance
Read through the README, run the synthea flexporter with something like `./run_synthea -fm qicore_130.yaml`. Look at the output patient file and confirm that the appropriate resources and fields are exported. You can create additional patients with the `-p {populationSize}` option. You can run resulting patients through fqm-execution calculations with CMS130 and make sure they calculate appropriately.

### DIFF
--- a/synthea/README.md
+++ b/synthea/README.md
@@ -1,0 +1,75 @@
+# Bulk Data Generation with Synthea
+
+The [Synthea](#https://github.com/synthetichealth/synthea) project can be used to generate realistic (but not real) patient data for use in a testing context. Synthea patient data can be exported according to the FHIR specification but may not be conformant to certain IGs and profiles. The Synthea [Flexporter](#https://github.com/synthetichealth/synthea/wiki/Flexporter) can be used to implement mappings during the export process that can add, remove, or update the patient data to be conformant to FHIR profiles.
+
+## Usage
+
+- Download the Synthea codebase from https://github.com/synthetichealth/synthea
+- In the Synthea project, use `./run_synthea -fm {mapping file location}` to run synthea with the flexporter
+- You may also use the flexporter standalone to map an existing exported file `./run_flexporter -fm {mapping_file_location} -s {source_fhir_file}`
+- See the [flexporter documentation](#https://github.com/synthetichealth/synthea/wiki/Flexporter) for additional information on the flexporter, mapping file, and limitations.
+
+## Building the Mapping File
+Quality Measurement calculation requires data conformant with qicore, and expansive IG, which would require expansive effort to fully map. As such, we can piecemeal address the IG requirements by supporting requirements for individual measures. The recommended process for creating a mapping that addresses a set of measures is:
+
+1. Use [elm-parser-for-ecqms fhir_review branch](#https://github.com/projecttacoma/elm-parser-for-ecqms/tree/fhir_review) and get data requirements to build a combined list of mustSupports for all resources across the set of measures.
+2. Use [fqm-execution](#https://github.com/projecttacoma/fqm-execution) and get data requirements to build a full list of profiles used.
+3. For all resource types, ensure resource is exported as a top level resource by Synthea or already supported by the flexporter mapping. If not, use the flexporter `create_resource` action to export the resource based on a logical existing exported resource or based on a logical Synthea module state. Resources that are only used for SDEs may be initially ignored.
+3. Make sure all profiles are applied to the correct exported resource.
+4. For each must support:
+    - Check the resource's [qicore 4.1.1](#https://hl7.org/fhir/us/qicore/STU4.1.1/) profile Snapshot Table to check if the mustSupport is also required by the profile (minimum cardinality 1). If not, it may be initially ignored.
+    - If required, check if synthea already exports it by looking in the [FhirR4.java](#https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/export/FhirR4.java) file. You can look for a string like `Condition()` for an example of how the Fhir object is created and what fields are set on it.
+    - If Synthea doesn't export it, check if the existing mapping file already adds it through a mapping.
+    - If not, fill from something logical that synthea does export (if something logical exists). Understanding what values are logical may involve looking at the FHIR specification and/or the measure's CQL for how the must support field is used in the measure logic.
+    - If a logical field doesn't exist, fill from a chosen pre-set value or choose randomly from a limited set of values 
+
+## CMS130 Mapping Example
+
+Below is the example list of resources and must support elements collated from elm-parser-for-ecqms data requirement calculation. Unexported values are marked with a `*` with notes about how their export was resolved in the `qicore_130.yaml` mapping file.
+
+- Condition:
+  - "code",
+  - "clinicalStatus",
+  - "onset",
+  - "abatement"
+- Coverage: * exported as a contained resource only -> ignored since Coverage is only used for SDEs
+  - "type",
+  - "period"
+- DeviceRequest: * not exported -> created based on existing exported Device resources
+  - "code",
+  - "authoredOn",
+  - "status",
+  - "intent",
+  - "modifierExtension", * not exported -> ignored since modifierExtension is only used for doNotPerform
+  - "modifierExtension.url",
+  - "modifierExtension.value"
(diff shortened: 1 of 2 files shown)


hossenlopp (changes requested):
For some reason running with this mapping causes the output to not have the `id` fields on many of the resources. This causes issues when running with fqm-execution as the `id`s are necessary for proper execution.

lmd59:
> For some reason running with this mapping causes the output to not have the `id` fields on many of the resources. This causes issues when running with fqm-execution as the `id`s are necessary for proper execution.

This is a bug in flexporter, which is fixed by this PR: https://github.com/synthetichealth/synthea/pull/1501.

elsaperelli (changes requested):
I could be misunderstanding the purpose of must supports, but there were some must supports that were not present on the generated patient's resources.

Other than that, just some small feedback!

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Building the Mapping File
+Quality Measurement calculation requires data conformant with qicore, and expansive IG, which would require expansive effort to fully map. As such, we can piecemeal address the IG requirements by supporting requirements for individual measures. The recommended process for creating a mapping that addresses a set of measures is:
+
+1. Use [elm-parser-for-ecqms fhir_review branch](https://github.com/projecttacoma/elm-parser-for-ecqms/tree/fhir_review) and get data requirements to build a combined list of mustSupports for all resources across the set of measures.
+2. Use [fqm-execution](https://github.com/projecttacoma/fqm-execution) and get data requirements to build a full list of profiles used.
+3. For all resource types, ensure resource is exported as a top level resource by Synthea or already supported by the flexporter mapping. If not, use the flexporter `create_resource` action to export the resource based on a logical existing exported resource or based on a logical Synthea module state. Resources that are only used for SDEs may be initially ignored.
+3. Make sure all profiles are applied to the correct exported resource.
```
lmd59, line 18:
```suggestion
4. Make sure all profiles are applied to the correct exported resource.
```

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## CMS130 Mapping Example
+
+Below is the example list of resources and must support elements collated from elm-parser-for-ecqms data requirement calculation. Unexported values are marked with a `*` with notes about how their export was resolved in the `qicore_130.yaml` mapping file.
+
+- Condition:
+  - "code",
+  - "clinicalStatus",
+  - "onset",
+  - "abatement"
```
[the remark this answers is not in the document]
lmd59: Synthea does export abatement, but only if the modules cause the condition to end (i.e. abate). Seems like that didn't happen in the runs you did - it may be a less common module state to end the condition.
Here's how synthea does it: https://github.com/synthetichealth/synthea/blob/055d977ec40f52cd5d386395297930241ad4e34c/src/main/java/org/mitre/synthea/export/FhirR4.java#L1651

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+3. Make sure all profiles are applied to the correct exported resource.
+4. For each must support:
+    - Check the resource's [qicore 4.1.1](https://hl7.org/fhir/us/qicore/STU4.1.1/) profile Snapshot Table to check if the mustSupport is also required by the profile (minimum cardinality 1). If not, it may be initially ignored.
+    - If required, check if synthea already exports it by looking in the [FhirR4.java](https://github.com/synthetichealth/synthea/blob/master/src/main/java/org/mitre/synthea/export/FhirR4.java) file. You can look for a string like `Condition()` for an example of how the Fhir object is created and what fields are set on it.
+    - If Synthea doesn't export it, check if the existing mapping file already adds it through a mapping.
+    - If not, fill from something logical that synthea does export (if something logical exists). Understanding what values are logical may involve looking at the FHIR specification and/or the measure's CQL for how the must support field is used in the measure logic.
+    - If a logical field doesn't exist, fill from a chosen pre-set value or choose randomly from a limited set of values 
+
+## CMS130 Mapping Example
+
```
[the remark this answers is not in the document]
lmd59: Not really here - this is from elm-parser-for-ecqms because we don't have must supports working for fqm-execution. But elm-parser-for-ecqms doesn't list profiles, so to get the right profile urls, I checked them against fqm-execution data requirements.

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+  - "modifierExtension.value"
+- Encounter:
+  - "status",
+  - "type",
+  - "period",
+  - "diagnosis",*
+   - "diagnosis.condition", * not exported -> set using a js function that finds conditions that reference this encounter and creates a reference to that condition
+  - "hospitalization"
+- MedicationRequest:
+  - "medication",
```
elsaperelli, line 53: So this technically supports both `medicationCodeableConcept` and `medicationReference` right? Do we want to include the `[x]` here? Actually I am not totally sure Synthea supports `medicationReference` ?

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+  - "doNotPerform", * not exported -> set to false because all resources exported from Synthea are logically performed
+  - "status",
+  - "intent",
+  - "dosageInstruction",
+  - "dispenseRequest", * not exported -> set from the `authoredOn` field because the measure logic coalesces `dispenseRequest.validityPeriod.start` with `authoredOn`
+  - "authoredOn"
+- Observation:
+  - "code",
+  - "value",
+  - "effective",
```
elsaperelli, line 63: Same sort of question as above- do we care whether Synthea does or doesn't export all of these choice types? Do we want to indicate they are choice types?

synthea/README.md
```diff
@@ -0,0 +1,75 @@
...
+  - "status",
+  - "category"
+- Procedure:
+  - "code",
+  - "performed",
+  - "status"
+- ServiceRequest: * exported as a contained resource only -> created based on existing exported Procedure resources
```
elsaperelli, line 70: Is there a reason we don't include some of the profiles included in the `yaml` mapping file like Immunization and Claim?

synthea/README.md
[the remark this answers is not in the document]
elsaperelli: Gotcha, was wondering if we should include that in this description but I think it's good as is!

elsaperelli (changes requested):

synthea/README.md
lmd59: This is just a list from CMS130's data requirements. I believe Immunization and Claim are not required for CMS130.

synthea/README.md
lmd59: The list here is taken directly from the must supports, which don't list a specific choice type. We don't necessarily need full coverage across all possible things that the measure could calculate against. Our main goal is to produce data that measure calculation won't choke on.

synthea/README.md
```diff
@@ -0,0 +1,75 @@
+# Bulk Data Generation with Synthea
+
+The [Synthea](https://github.com/synthetichealth/synthea) project can be used to generate realistic (but not real) patient data for use in a testing context. Synthea patient data can be exported according to the FHIR specification but may not be conformant to certain IGs and profiles. The Synthea [Flexporter](https://github.com/synthetichealth/synthea/wiki/Flexporter) can be used to implement mappings during the export process that can add, remove, or update the patient data to be conformant to FHIR profiles.
+
+## Usage
+
+- Download the Synthea codebase from https://github.com/synthetichealth/synthea
+- In the Synthea project, use `./run_synthea -fm {mapping file location}` to run synthea with the flexporter
```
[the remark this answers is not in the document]
hossenlopp: It was just merged so this can be removed.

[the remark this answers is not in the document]
lmd59: Done!

hossenlopp (approved):

PR #52, opened 2024-11-06
Add QPP profiles to flexporter mapping file

elsaperelli:
# Summary
This PR adds profiles defined by the generated CMS Quality Payment Program (QPP) IG to the existing qicore flexporter mapping file. 

## New behavior
- New QI Core profiles defined: `AdverseEvent`, `Communication`, `MedicationAdministration`, `MedicationDispense` and `Task`.

## Code changes
- `synthea/qpp_qicore.yaml` - renamed file and added additional QI Core profile creation
- `synthea/README.md` - updates to explain the changes to the mapping file

# Testing guidance
- On a local instance of [synthea](https://github.com/synthetichealth/synthea), uncomment line 301 of `src/main/resources/synthea.properties` in order to use the terminology server.
- In the synthea root directory, run `./run_synthea -fm <path to qpp_qicore.yaml>`
- Inspect the resources of the generated patient in the `/output` directory


synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,171 @@
...
+                  reference: `Condition/${conditionEntry.resource.id}`
+                }
+              }];
+            }
+          }
+  - name: Create Resources
+    create_resource:
+      - resourceType: ServiceRequest
+        based_on:
+          resource: Procedure.performed.ofType(Period) # handle value setting for period choice type
```
dehall, line 67: This can be cleaned up now - since Lauren originally wrote the earlier mapping file I added support for `if` -- see https://github.com/synthetichealth/synthea/pull/1525
Basically instead of duplicated create_resource entries based on different possible types of Procedure.performed, you can collapse that down to one overall entry for Procedure, and just different fields with an `if` that distinguishes between the possible field types

elsaperelli: OH awesome, thanks Dylan!!

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+            value: order
+            # Not sure what AdverseEvent would be close to, creating one per patient
+      - resourceType: AdverseEvent
+        based_on:
+          resource: Patient
+          profiles:
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-adverseevent
+        fields:
+          - location: AdverseEvent.event.coding
+            value: $randomCode([http://hl7.org/fhir/ValueSet/adverse-event-type])
```
lmd59, line 110: We run the risk of creating unrealistic data when we select a random code. For this code set especially, a lot of the codes indicate a death, which may be problematic when the patient doesn't have other data indicating death. We can probably leave it for now and see how it turns out in testing as it may not actually be a big deal. If we want to be extra proactive, we could identify some common bad outcomes that come from synthea modules and create the AdverseEvent resource based on that. (I would guess that Synthea already endeavors to create some adverse events even if they're not labeled as such, and Dylan might have more insight into this.)

lmd59 (commented):
Cleanup looks good, and the new resources are all working! Yay! A couple of comments about potentially making the data more relevant if we can do it easily enough. And a quick question: For resources that were already exported by synthea or the previous 130 mapping file, were there any fields from QPP that weren't already exported on those resources? Wondering in order to contextualize the upcoming task.

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+      - resourceType: Communication
+        based_on:
+          resource: Patient
+          profiles:
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-communication
+        fields:
+          - location: Communication.status
+            value: completed
+      # Adding a MedicationAdministration for each Procedure resource
+      - resourceType: MedicationAdministration
```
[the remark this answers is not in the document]
elsaperelli: I did! I wasn't totally sure if that made sense, but if you think it does then I am happy to change that.

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-medicationadministration
+        fields:
+          - location: MedicationAdministration.status
+            value: completed
+          - location: MedicationAdministration.medication.coding
+            value: $randomCode([http://hl7.org/fhir/ValueSet/medication-codes])
+          - location: MedicationAdministration.effective
+            value: $getField([Procedure.performed])
+      # Adding a MedicationDispense for each Procedure resource
+      - resourceType: MedicationDispense
```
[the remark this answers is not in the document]
elsaperelli: I did! I wasn't totally sure if that made sense, but if you think it does then I am happy to change that.

synthea/qpp_qicore.yaml
elsaperelli: Sounds good! Or, I am happy to not use `randomCode` here if it would make more sense to just pick a code from that list...but then we wouldn't have as much variability so I am open to either.

elsaperelli:
> Cleanup looks good, and the new resources are all working! Yay! A couple of comments about potentially making the data more relevant if we can do it easily enough. And a quick question: For resources that were already exported by synthea or the previous 130 mapping file, were there any fields from QPP that weren't already exported on those resources? Wondering in order to contextualize the upcoming task.

To answer your question, no, I did not find any fields missing from the resources that were already exported by synthea or the previous 130 mapping file.

Regarding the status codes for some of the profiles that we have arbitrarily set in the mapping file, is it ok for me to go ahead and use `randomCode` for them now that Dylan added that functionality to Synthea?

synthea/qpp_qicore.yaml
dehall: Off the top of my head I can't think of any adverse events in the modules, other than possibly some of the surgeries have a risk of death. I can look for some though if that's of interest.

This also makes me think there should be some way of saying "random code from these specific few codes" rather than requiring the use of a terminology server and a published ValueSet. I can probably get that implemented pretty quickly

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+      # Creating one MedicationAdministration for each MedicationRequest
+      - resourceType: MedicationAdministration
+        based_on:
+          resource: MedicationRequest
+          profiles:
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-medicationadministration
+        fields:
+          - location: MedicationAdministration.status
+            value: completed
+          - location: MedicationAdministration.medication.coding
```
lmd59, line 129: The MedicationAdministration.medication should probably be set to the MedicationRequest.medication

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+      # Creating one MedicationDispense for each MedicationRequest resource
+      - resourceType: MedicationDispense
+        based_on:
+          resource: MedicationRequest
+          profiles:
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-medicationdispense
+        fields:
+          - location: MedicationDispense.status
+            value: $randomCode([http://hl7.org/fhir/ValueSet/medicationdispense-status,code])
+          - location: MedicationDispense.medication.coding
```
lmd59, line 142: The MedicationDispense.medication should probably be set to the MedicationRequest.medication

synthea/qpp_qicore.yaml
```diff
@@ -0,0 +1,156 @@
...
+            value: $getField([MedicationRequest.dispenseRequest.validityPeriod])
+      # Creating one MedicationDispense for each MedicationRequest resource
+      - resourceType: MedicationDispense
+        based_on:
+          resource: MedicationRequest
+          profiles:
+            - http://hl7.org/fhir/us/qicore/StructureDefinition/qicore-medicationdispense
+        fields:
+          - location: MedicationDispense.status
+            value: $randomCode([http://hl7.org/fhir/ValueSet/medicationdispense-status,code])
```
lmd59, line 141: Just a note for the future work if selecting from a list of codes becomes an available feature... because the MedicationRequest will only be `active` or `stopped`, we could potentially map (using the new `if` feature) `active` to the positive status values for dispense (preparation, in-progress, on-hold, completed) and map `stopped` to the negative values (cancelled, entered-in-error, stopped, declined).

lmd59 (commented):
Just a recommendation for the `.medication` fields and a note for future tasking!

lmd59 (approved):
Looks good! Nice updates!

PR #56, opened 2025-02-25
Use fhir-spec-tools in place of previous scripts

elsaperelli:
# Summary
This PR adds [fhir-spec-tools](https://github.com/projecttacoma/fhir-spec-tools) as a dependency in order to replace previous scripts and script outputs that fhir-spec-tools was intended to replace. 

## New behavior
There should be no new behavior

## Code changes
- `package.json` / `package-lock.json` - add `fhir-spec-tools` as a dependency, remove "parse-compartment-def" command
- `src/compartment-definition` - removed `choice-types.json`, `compartmentdefinition-patient.json`, `mandatory-elements.json`, `patient-references.json`, `patientExportResourceTypes.json` - output files from scripts that are no longer needed since they were replaced by fhir-spec-tools
- `src/scripts/getChoiceTypesForResource.js` / `src/scripts/parseCompartmentDefinition.js` / `src/scripts/parseStructureDefinitions.js` - remove scripts that are no longer needed
- `src/services/export.service.js` - replace `patientExportResourceTypes.json` with `patientAttributePaths` export from fhir-spec-tools. Since the script in fhir-spec-tools no longer outputs just the list of resource types, we need to get the keys from the `patientAttributePaths` object which will give us the patientResourceTypes that we need
- `src/util/exportToNDJson.js` - replace compartment-definition files with fhir-spec-tools exports

# Testing guidance
- `npm run check`
- `npm run start`
- Check different endpoints to make sure they all work as expected!


hossenlopp (approved):
looks good!

zacharyrobin (approved):
Tested and works on my end! Looks great :)

PR #57, opened 2025-02-26
Criteria groups

elsaperelli:
# Summary
Functionality to export criteria-based groups, which are defined by the group member-filters

## New behavior
For a non-actual group that's been created, the server can now do a group export that will actualized the members of that group based on the member-filters as defined in https://build.fhir.org/ig/HL7/bulk-data/branches/argo24/group.html#bulk-cohort-api

Limitations: Our server does not have enough support for date searches that we can demonstrate this for a group directly built from a measure's IPP functionality. We already have ticket 1540 created to address our date limitations. For now, we need custom groups to best test this functionality.

## Code changes

- Adds lodash library for easier set management
- Updates in `export.service.js` to actualize the members of the group if the group is not `actual=true`
- Updates in exportToNDJson.js to move some typeFilter logic into a reusable `addTypeFilter` function
- `groupUtils.js` contains the main meat of the logic for this functionality. It has a couple of small changes to `verifyPatientsInGroup` to compensate for non-actual groups. Then `actualizeGroup` creates the group of Patients based on the member-filter expressions

# Testing guidance

Because of our date limitations, I've created a doctored group to test this functionality. I will share an Insomnia set in slack.

- `npm run check`
- `npm run db:reset`
- `npm run start`
- Run POSTs from the insomnia set
- Copy the group id to the Group export GET
- Check the resulting status
- Check the resulting Patient ndjson. This file should have the 3 test patients from the cervical cancer screening measure, but not the 4th patient that was uploaded in the second transaction bundle. This 4th patient is a copy of the denominator patient but with an altered code that should exclude it from the defined group.
- (optional) `npm run db:reset` and `npm run upload-bundles` if you want to return to a more populated database

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Alternatively, a POST request (`POST [fhir base]/Patient/$export`) can be sent. 
 
 #### All Patients in a Group
 
-FHIR Operation to obtain a detailed set of FHIR resources of diverse resource types pertaining to all patients that belong to a defined Group resource.
+FHIR Operation to obtain a detailed set of FHIR resources of diverse resource types pertaining to all patients that belong to a defined Group resource. In addition to an actualized group with specifically defined membership, this can also be applied to a criteria based group (as defined in the [Bulk Data Access IG](https://build.fhir.org/ig/HL7/bulk-data/branches/argo24/group.html#bulk-cohort-api)), which uses expressions to define membership. For a criteria based group, the group membership will be actualized at the time of `$export`.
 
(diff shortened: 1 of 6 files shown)


elsaperelli (commented):
Looks and is working well! Thanks for that Insomnia collection. My only suggestion besides these two small comments is perhaps adding a bit about this to the README? Really just so this is highlighted in this repo, but also okay if we don't want to add it quite yet.

elsaperelli (commented):
🥳

lmd59 (approved):

PR #66, opened 2026-03-30
Update CI to use node 24. Dockerfile now using hardened images

hossenlopp:
# Summary

Update to use newer node for CI and Docker builds. Additional address many out of date npm audit concerns. Updated ESlint to the latest version.

## New behavior

None

## Code changes

None

# Testing guidance

- Run in docker with `docker compose up --build`.
- `npm run check`


elsaperelli:
It looks like I am still getting some npm audit issues when I run `npm install`

lmd59 (commented):
Did some testing with data loaded into the docker instance, and that was looking good to me. Changes also seem to look good!
If npm audit issues are in scope, then I agree that I'm seeing some issues. `npm audit fix` fixes some, but can't address the critical I'm seeing.
I did a little poking at the critical issue, which seemed to be coming from a sequelize dependency. This is a dependency out of @asymmetrik/fhir-qb, which hasn't been updated in quite a long time because it has actually been moved into node-fhir-server-core, which has much more recent updates. Dependency being used is the QueryBuilder class, which can be found in core here: https://github.com/bluehalo/node-fhir-server-core/blob/75a1011e3cfdffa76a3184ddf57921843126c60b/packages/fhir-qb/index.js#L45 
Shouldn't be too hard a replacement.

lmd59 (commented):
Changes are looking good. I noticed if I run `npm audit fix` on branch, it fixes one moderate vulnerability. Any reason not to include that last fix?
Additionally, is it worth a little update on the PR description just to capture a little more of what's happening here in case we're looking at historical changes in the future?

lmd59 (approved):
👍 🎉

PR #67, opened 2026-04-01
Collect data

lmd59:
# Summary
Adds support for a basic $collect-data according to the [STU5 definition](https://hl7.org/fhir/us/davinci-deqm/STU5/OperationDefinition-collect-data.html)

## New behavior
Supports $collect-data parameters periodStart, periodEnd, measureId, and subject. Does not support all STU5 parameters or alternative parameters/requirements in the [latestBuild](https://build.fhir.org/ig/HL7/davinci-deqm/en/OperationDefinition-collect-data.html)

## Code changes

- Update premadeBundle upload default
- Add routes to app.js
- Add main collectData endpoint handling and validation to export service
- Create collectDataUtils for utilities: patient bundle and measure report creation, finding patient resources, creating typeFilters based on effective data requirements
- Small update to exportToNDJson.js to increase robustness
- Small updates to serviceUtils to handle gathering parameter input
- Testing: Add and update fixtures and unit testing of $collect-data operation in export.service.test.js

# Testing guidance

- `npm run check`
- Add ecqm-content-qicore-2025 bundles to your bulk-export-server directory (will be used instead of 2021 bundles)
- `npm run upload-bundles` (should load new default 2025 bundles)
- `npm run start`
- Run POST (or GET w/ URL parameters) to http://localhost:3000/Measure/$collect-data with
```
{
  "resourceType": "Parameters",
  "parameter": [
    {"name": "periodStart",
		 "valueDate": "2025-01-01"
		},
    {"name": "periodEnd",
		 "valueDate": "2025-12-31"
		},
		{
      "name": "measureId",
			"valueId": "CMS125FHIRBreastCancerScreening"
    },
		{
      "name": "measureId",
			"valueId": "CMS133FHIRCataracts2040BCVAwithin90Days"
    },
    {
      "name": "subject",
      "valueString": "Patient/62ea0c3d-46da-48a1-87dd-d1927ed2df75"
    }
  ]
}
```
- Experiment with other inputs, but see CollectDataBrainstorming document for caveats and potential follow-up actions (bottom of document) that would make this implementation more complete

### DIFF
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -450,11 +450,10 @@ const collectData = async (request, reply) => {
     const bundles = await Promise.all(
       patientIds.map(async id => {
         const patient = await findResourceById(id, 'Patient');
-        const measureReports = [];
-        const patientResourcePromises = measures.map(async measure => {
-          const patientResources = await findPatientResources(patient, measure);
-          measureReports.push(
-            createDataExchangeMeasureReport(
+        const resourcesMRPairs = await Promise.all(
+          measures.map(async measure => {
+            const patientResources = await findPatientResources(patient, measure);
+            const measureReport = createDataExchangeMeasureReport(
               measure,
               {
                 start: parameters.periodStart,
@@ -462,11 +461,11 @@ const collectData = async (request, reply) => {
               },
               id,
               patientResources
-            )
-          );
-          return patientResources;
-        });
-        const patientResourcesArray = await Promise.all(patientResourcePromises);
+            );
+            return [patientResources, measureReport];
+          })
+        );
+        const [patientResourcesArray, measureReports] = _.unzip(resourcesMRPairs);
         const uniqueResources = _.uniqBy(
           patientResourcesArray.flat(),
           resource => `${resource.resourceType}/${resource.id}`
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -6,6 +6,7 @@ const patientResourceTypes = Object.keys(patientAttributePaths);
 const { createOperationOutcome } = require('../util/errorUtils');
 const { verifyPatientsInGroup, actualizeGroup } = require('../util/groupUtils');
 const { gatherParams } = require('../util/serviceUtils');
+const _ = require('lodash');
 const {
   createDataExchangeMeasureReport,
   createPatientBundle,
@@ -459,33 +460,42 @@ const collectData = async (request, reply) => {
     // TODO: measureId (0..*) should be handled correctly when gathering parameters, add other 0..* are handled as arrays in gatherParams
     // i.e., could have additional {"name": "measureId","valueId": "measure1"} specified
 
-    const measureId = parameters.measureId[0];
-    const measure = await findResourceById(measureId, 'Measure');
-    if (!measure) {
-      throw Error(`Unable to find measure with measureId ${measureId}`);
-    }
     const patientIds = [parameters.subject.split('Patient/')[1]];
+    // Check for measure resolution - errors if there are any issues with measures passed
+    const measurePromises = parameters.measureId.map(async id => {
+      const measure = await findResourceById(id, 'Measure');
+      if (!measure) {
+        reply.code(404).send(new Error(`Unable to find measure with measureId ${id}`));
+      }
+      return measure;
+    });
+    const measures = await Promise.all(measurePromises);
 
     const bundles = await Promise.all(
       patientIds.map(async id => {
         const patient = await findResourceById(id, 'Patient');
-        const patientResources = await findPatientResources(patient, measure);
-        return createPatientBundle(
-          patient,
-          patientResources.map(r => {
-            return { resource: r };
-          }), //map from resources to BundleEntrys
-          patient.fullUrl ?? `urn:uuid:${id}`,
-          createDataExchangeMeasureReport(
-            measure,
-            {
-              start: parameters.periodStart,
-              end: parameters.periodEnd
-            },
-            id,
-            patientResources
-          )
+        const measureReports = [];
+        const patientResourcePromises = measures.map(async measure => {
+          const patientResources = await findPatientResources(patient, measure);
+          measureReports.push(
+            createDataExchangeMeasureReport(
+              measure,
+              {
+                start: parameters.periodStart,
+                end: parameters.periodEnd
+              },
+              id,
+              patientResources
+            )
+          );
+          return patientResources;
+        });
+        const patientResourcesArray = await Promise.all(patientResourcePromises);
+        const uniqueResources = _.uniqBy(
+          patientResourcesArray.flat(),
+          resource => `${resource.resourceType}/${resource.id}`
         );
+        return createPatientBundle(patient, uniqueResources, measureReports);
       })
     );
 
--- a/src/services/export.service.js
+++ b/src/services/export.service.js
@@ -432,33 +432,9 @@ async function validatePatientReferences(patientParam, reply) {
 const collectData = async (request, reply) => {
   const parameters = gatherParams(request.method, request.query, request.body, reply);
 
-  // TODO: make sure measureId isn't specified differently in url and parameters
+  // TODO: validate measureId isn't specified differently in url and parameters
   if (validateCollectDataParams(parameters, reply)) {
     request.log.info('Measure >>> $collect-data');
-
-    // Easiest case (measureId and single patient subject) - TODO: handle other parameters, multiple patients/measures
-    //   Example:
-    //   {
-    //   "resourceType": "Parameters",
-    //   "parameter": [
-    //     {"name": "periodStart",
-    // 		 "valueDate": "2023-01-01"
-    // 		},
-    //     {"name": "periodEnd",
-    // 		 "valueDate": "2023-12-31"
-    // 		},
-    // 		{
-    //       "name": "measureId",
-    // 			"valueId": "measure1"
-    //     },
-    //     {
-    //       "name": "subject",
-    //       "valueString": "Patient/patient03"
-    //     }
-    //   ]
-    // }
-    // TODO: measureId (0..*) should be handled correctly when gathering parameters, add other 0..* are handled as arrays in gatherParams
-    // i.e., could have additional {"name": "measureId","valueId": "measure1"} specified
 
     const patientIds = [parameters.subject.split('Patient/')[1]];
     // Check for measure resolution - errors if there are any issues with measures passed
@@ -510,7 +486,7 @@ const collectData = async (request, reply) => {
  * @param {Object} reply the response object
  */
 function validateCollectDataParams(parameters, reply) {
-  // TODO: Update with additional validations (and not supporteds as applicable)
+  // TODO: Update with additional validations
 
   let unrecognizedParams = [];
   Object.keys(parameters).forEach(param => {
@@ -529,7 +505,8 @@ function validateCollectDataParams(parameters, reply) {
         'lastReceivedOn',
         'organizationResource',
         'organization',
-        'validateResources'
+        'validateResources',
+        'dataEndpoint'
       ].includes(param)
     ) {
       unrecognizedParams.push(param);
@@ -542,6 +519,24 @@ function validateCollectDataParams(parameters, reply) {
         createOperationOutcome(
           `The following parameters are unrecognized by the server: ${unrecognizedParams.join(', ')}.`,
           { issueCode: 400, severity: 'error' }
+        )
+      );
+    return false;
+  }
+  // TODO: support other parameters that are currently unsupported
+  let unsupportedParams = [];
+  Object.keys(parameters).forEach(param => {
+    if (!['periodStart', 'periodEnd', 'measureId', 'subject'].includes(param)) {
+      unsupportedParams.push(param);
+    }
+  });
+  if (unsupportedParams.length > 0) {
+    reply
+      .code(501)
+      .send(
+        createOperationOutcome(
+          `The following parameters are not yet supported by the server: ${unsupportedParams.join(', ')}.`,
+          { issueCode: 501, severity: 'error' }
         )
       );
     return false;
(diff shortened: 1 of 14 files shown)


src/util/collectDataUtils.js
```diff
@@ -0,0 +1,186 @@
...
+const patientResourceTypes = Object.keys(patientAttributePaths);
+const { addTypeFilter, getDocuments } = require('./exportToNDJson');
+const _ = require('lodash');
+
+/**
+ * Creates a patient bundle resource. Creates using a patient resource and
+ * an array of the patient's associated resources
+ * @param {Object} patient FHIR Patient object
+ * @param {Array} resources array of resources associated with the patient
+ * @param {Array} measureReports array of MeasureReport specifying measure information for data exchange
```
elsaperelli:
```suggestion
 * @param {Array} measureReports array of MeasureReport resources specifying measure information for data exchange
```

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,186 @@
...
+      fullUrl: measureReport.fullUrl ?? `urn:uuid:${measureReport.id}`
+    });
+  });
+
+  return bundle;
+}
+
+/**
+ * Creates a FHIR data exchange MeasureReport from measure and subject data
+ * https://build.fhir.org/ig/HL7/davinci-deqm/StructureDefinition-datax-measurereport-deqm.html
```
elsaperelli: For some reason this link isn't working for me? But I was thinking we could start getting in the habit of using version specific links anyways so we know in the future!

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,186 @@
...
+  // month is 0 indexed
+  const month = date.getMonth() + 1;
+  const day = date.getDate();
+  return `${year}-${month < 10 ? `0${month}` : month}-${day < 10 ? `0${day}` : day}`;
+}
+
+/**
+ * Finds all resources related to the passed patient resource but limited by the data requirements on the passed measure
+ * @param {Object} patient fhir patient resource
+ * @param {Object} measure fhir measure resource
```
elsaperelli: Can we use FHIR.Patient and FHIR.Measure as the types here?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,186 @@
...
+  const [patientTypes, nonPatientTypes] = types.reduce(
+    ([patient, nonPatient], type) => {
+      (patientResourceTypes.includes(type) ? patient : nonPatient).push(type);
+      return [patient, nonPatient];
+    },
+    [[], []]
+  );
+  // TODO: Handle non-patient types more elegantly/completely
+  // i.e. Patient references MedicationRequest references Medication could end up with Medication as non-patient type
+  console.warn('Ignoring non-patient types found in data requirements:', nonPatientTypes);
```
elsaperelli:
```suggestion
  console.warn('Ignoring non-patient types found in data requirements: ', nonPatientTypes);
```

elsaperelli: Do we need to format `nonPatientTypes` at all?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,186 @@
...
+    }); // potential multiple codes are comma-separated to be ORed for this path
+    const tfStr = `${dr.type}?${fhirQueries?.join('&')}`; //Example value: 'Procedure?code=1,2&category=3,4'
+    if (typeFilters[dr.type]) {
+      typeFilters[dr.type]?.push(tfStr);
+    } else {
+      typeFilters[dr.type] = [tfStr];
+    }
+  });
+
+  // array of typeFilters that should be treated as OR'd (empty array will be ignored and flattened)
```
elsaperelli, line 171:
```suggestion
  // array of typeFilters that should be treated as ORed (empty array will be ignored and flattened)
```
Or change the other ORed to OR'd for consistency

src/services/export.service.js
```diff
@@ -417,4 +423,123 @@ async function validatePatientReferences(patientParam, reply) {
...
+              {
+                start: parameters.periodStart,
+                end: parameters.periodEnd
+              },
+              id,
+              patientResources
+            )
+          );
+          return patientResources;
+        });
```
elsaperelli: I don't have a crazy strong opinion on this, but would like to discuss. Something about the `measureReports.push` with `patientResources` _within_ the `.map` just kind of confuses me a bit. I see how it works, but wondering if there would be a prettier way to do this? Could we maybe return `patientResources` and that `createDataExchangeMeasureReport` output and then map it after the `Promise.all()`?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+  return {
+    resourceType: 'MeasureReport',
+    id: uuidv4(),
+    measure: measure.url?.includes('|') ? measure.url : `${measure.url}|${measure.version}`, //canonical measure/version
+    period: measurementPeriod,
+    status: 'complete',
+    type: 'data-collection',
+    subject: { reference: `Patient/${subjectId}` },
+    date: new Date().toISOString(),
+    reporter: { reference: 'Organization/bulk-export-server' }, //TODO: do we need to send an organization resource?
```
elsaperelli, line 66: It seems like we do? But this seems like a good enough approach for now.

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+    extension: [
+      {
+        url: 'http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/extension-submitDataUpdateType',
+        valueCode: 'snapshot'
+      }
+    ],
+    evaluatedResource: patientResources?.map(r => {
+      return { reference: `${r.resourceType}/${r.id}` };
+    }),
+    contained: [{ resourceType: 'Organization', id: 'bulk-export-server' }]
```
elsaperelli, line 79: Just curious about where this came from? I don't think I see it in the data exchange MeasureReport profile?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+      (patientResourceTypes.includes(type) ? patient : nonPatient).push(type);
+      return [patient, nonPatient];
+    },
+    [[], []]
+  );
+  // nonPatientTypes can be in data requirements when they may be referenced from other resources that reference patients
+  // i.e. Patient references MedicationRequest references Medication could end up with Medication as non-patient type
+  // ignored for now, but should either get all (non-patient query) or be able to do more complex query creation
+  if (nonPatientTypes.length > 0) {
+    console.warn('Ignoring non-patient types found in data requirements:', nonPatientTypes);
```
elsaperelli, line 103: Do we want to format the `nonPatientTypes` array of strings here?

elsaperelli, line 103:
```suggestion
    console.warn('Ignoring non-patient types found in data requirements: ', nonPatientTypes);
```

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+      typeFilters[dr.type] = [];
+      return;
+    }
+    //all codefilters have a path and proper code and can be added to our typefilter array
+    const fhirQueries = dr.codeFilter?.map(cf => {
+      const hasVS = cf.path && cf.valueSet;
+      const hasCode = cf.path && cf.code && cf.code.every(coding => !!coding.code);
+      if (hasVS && hasCode) {
+        // default to valueset method for now, but this should probably be expanded to a full list of codes from the vs with additional
+        // codes appended (or separated into separate top-leve queries where other included codeFilters are repeated
```
elsaperelli:
```suggestion
        // codes appended (or separated into separate top-level queries where other included codeFilters are repeated
```

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+        // Example: 'Procedure?code=1&category=3,4','Procedure?code=1&category:in=vs)')
```
elsaperelli:
```suggestion
        // Example: 'Procedure?code=1&category=3,4','Procedure?code=1&category:in=vs')
```
I think?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+        return `${cf.path}:in=${cf.valueSet}`;
+      } else if (hasVS) {
+        return `${cf.path}:in=${cf.valueSet}`;
```
elsaperelli, line 156: Are these returns the same?

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+      } else {
+        // hasCode
+        // potential multiple codes are comma-separated to be ORed for this path
+        return `${cf.path}=${cf.code?.map(coding => coding.code).join(',')}`;
+      }
+    });
+    const tfStr = `${dr.type}?${fhirQueries?.join('&')}`; //Example value: 'Procedure?code=1,2&category=3,4'
+    if (typeFilters[dr.type]) {
+      typeFilters[dr.type]?.push(tfStr);
```
elsaperelli:
```suggestion
      typeFilters[dr.type].push(tfStr);
```
I think this is safe to say? Maybe?

elsaperelli (changes requested):
Functionality is looking great!!! Some VERY minor comments :)

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+      fullUrl: measureReport.fullUrl ?? `urn:uuid:${measureReport.id}`
+    });
+  });
+
+  return bundle;
+}
+
+/**
+ * Creates a FHIR data exchange MeasureReport from measure and subject data
+ * https://build.fhir.org/ig/HL7/davinci-deqm/StructureDefinition-datax-measurereport-deqm.html
```
lmd59:
```suggestion
 * https://hl7.org/fhir/us/davinci-deqm/STU5/StructureDefinition-datax-measurereport-deqm.html
```

src/util/collectDataUtils.js
lmd59: This was an outstanding question from our work with submit-data, so I carried it over here. Doesn't seem like the most important thing to resolve at the moment. We do send the organization resource as part of the contained resources below for now because the receiving system is unlikely to already have that resource in our current implementation.

src/util/collectDataUtils.js
lmd59: See above - the receiving system is unlikely to have the resource that's referenced as the `reporter` so we send it as a contained resource.

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+    evaluatedResource: patientResources?.map(r => {
+      return { reference: `${r.resourceType}/${r.id}` };
+    }),
+    contained: [{ resourceType: 'Organization', id: 'bulk-export-server' }]
+  };
+}
+
+/**
+ * Finds all resources related to the passed patient resource but limited by the data requirements on the passed measure
+ * @param {Object} patient fhir patient resource
```
lmd59:
```suggestion
 * @param {FHIR.Patient} patient fhir patient resource
```

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+ * @param {Object} measure fhir measure resource
```
lmd59:
```suggestion
 * @param {FHIR.Measure} measure fhir measure resource
```

src/util/collectDataUtils.js
lmd59: https://developer.mozilla.org/en-US/docs/Web/API/console/warn_static
Comma separated inputs to console.warn are printed with separation between them already.

lmd59: Expected output formatting from passing `nonPatientTypes` in here is something like: `["type1", "type2", "type3"]` which seems reasonably readable to me.

src/util/collectDataUtils.js
lmd59: Yes- there really needs to be a different approach for the above case, but for now, defaulting to the same as this case so that it at least does something somewhat useful.

src/util/collectDataUtils.js
```diff
@@ -0,0 +1,175 @@
...
+      if (hasVS && hasCode) {
+        // default to valueset method for now, but this should probably be expanded to a full list of codes from the vs with additional
+        // codes appended (or separated into separate top-leve queries where other included codeFilters are repeated
+        // Example: 'Procedure?code=1&category=3,4','Procedure?code=1&category:in=vs)')
+        return `${cf.path}:in=${cf.valueSet}`;
+      } else if (hasVS) {
+        return `${cf.path}:in=${cf.valueSet}`;
+      } else {
+        // hasCode
+        // potential multiple codes are comma-separated to be ORed for this path
```
lmd59:
```suggestion
        // potential multiple codes are comma-separated to be OR'd for this path
```

src/util/collectDataUtils.js
lmd59: I read `ORed` as "Order" for a sec, so I think I like `OR'd` better? Changed the above instead.

src/services/export.service.js
lmd59: I can change it to a forEach and have the forEach loop push to patientResources at the same time. That way there isn't a weird map side effect, but we're still keeping the same efficiency. Will make that change.

lmd59: Jk- forEach can't give you a set up promises to Promise.all on. Ended up mapping to patientResources/MeasureReport pairs and then separating out into separate arrays with lodash function.

src/util/collectDataUtils.js
lmd59: Expected output formatting from passing nonPatientTypes in here is something like: ["type1", "type2", "type3"] which seems reasonably readable to me.

src/util/collectDataUtils.js
elsaperelli: Lol wow TIL!!!

ecqm-content-qicore-2025
elsaperelli, line 1: Do we want this? Should we add this to the `.gitignore` file?

I did notice that we should at least add this to the `.prettierignore` file.

elsaperelli (changes requested):
One last comment about adding `ecqm-content-qicore-2025` to some .ignore files but otherwise it's looking good!!!

ecqm-content-qicore-2025
lmd59: Good call!

elsaperelli (approved):
Lgtm!!

PR #68, opened 2026-04-23
$collect-data measureUrl

hossenlopp:
# Summary

Replaces measureId parameter with canonical measureUrl parameter for $collect-data.

## New behavior

`measureUrl` is now expected to be used to reference measures. Different versions of a measure with the same url can exist and must be specified with the version in the canonical.

## Code changes

`mongo.controller.js` - Added findResourceByCanonical to search for resources by canonical.

# Testing guidance

- Run unit tests.
- Test $collect-data by referencing measures by canonical url. Example parameter body using a canonical measureUrl.
     ```
     {
	"resourceType": "Parameters",
	"parameter": [
		{
			"name": "periodStart",
			"valueDate": "2023-01-01"
		},
		{
			"name": "periodEnd",
			"valueDate": "2023-12-31"
		},
		{
			"name": "measureUrl",
			"valueCanonical": "https://madie.cms.gov/Measure/CMS145FHIRCADBetaBlockerTherapyPriorMIorLVSD|0.4.000"
		},
		{
			"name": "subject",
			"valueString": "Patient/f6dafa7e-15a9-4fec-baea-59d43201caf8"
		}
	   ]
     }
     ```
- Clone a measure in the database changing the version for the clone. Ensure that not providing the version in the measureUrl results in an appropriate error.

### DIFF
--- a/src/scripts/uploadPremadeBundles.js
+++ b/src/scripts/uploadPremadeBundles.js
@@ -72,7 +72,7 @@ async function main() {
     const bundlePath = path.resolve(process.argv[2]);
     try {
       if (!searchPattern) {
-        searchPattern = /-bundle.json$/;
+        searchPattern = /.json$/;
       }
       console.log(`Finding bundles in ${bundlePath}.`);
       getBundleFiles(bundlePath, searchPattern);
--- a/src/scripts/uploadPremadeBundles.js
+++ b/src/scripts/uploadPremadeBundles.js
@@ -72,7 +72,7 @@ async function main() {
     const bundlePath = path.resolve(process.argv[2]);
     try {
       if (!searchPattern) {
-        searchPattern = /.json$/;
+        searchPattern = /-bundle.json$/;
       }
       console.log(`Finding bundles in ${bundlePath}.`);
       getBundleFiles(bundlePath, searchPattern);
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -17,6 +17,11 @@ const {
 // set bodyLimit to 50mb
 function build(opts) {
   const app = fastify({ ...opts, bodyLimit: 50 * 1024 * 1024 });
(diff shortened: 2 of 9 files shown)


src/scripts/uploadPremadeBundles.js
```diff
@@ -72,7 +72,7 @@ async function main() {
     const bundlePath = path.resolve(process.argv[2]);
     try {
       if (!searchPattern) {
-        searchPattern = /.json$/;
+        searchPattern = /-bundle.json$/;
```
lmd59: Just since this wasn't described - looks like this change might have been updated to prevent group json (or any other non-measure, non-patient json files) from being identified?

lmd59 (commented):
I'm having some trouble testing which seems unrelated to your code, but this is looking great! The one recommendation I have: it looks like this doesn't enforce the inclusion of at least one measureUrl parameter (cardinality 1..*), which we should probably include at this time.

src/server/app.js
```diff
@@ -17,6 +17,11 @@ const {
 // set bodyLimit to 50mb
 function build(opts) {
   const app = fastify({ ...opts, bodyLimit: 50 * 1024 * 1024 });
+  app.addContentTypeParser(
```
lmd59, line 20: Great update here! 👍

src/scripts/uploadPremadeBundles.js
hossenlopp: Yep... this was an accidental change as I was trying to get things to load. Will remove this.

hossenlopp: Removed in [29cf066](https://github.com/projecttacoma/bulk-export-server/pull/68/commits/29cf066b705225d2b1d389bc92ff6c44a0820e4f).

lmd59 (approved):

PR #70, opened 2026-05-20
Validation

lmd59:
# Summary
Validation for $collect-data, subjectGroup and subject Group reference support, and various other small fixes

## New behavior

- $collect-data request sends appropriate HTTP code responses for invalid input
- Can do $collect-data operation for groups of patients

Some bug fixes:
- Returns a Parameters resource (rather than array of bundles)
- Uses `valueReference` for patient references (for $collect-data subject but also for $export patient parameter)
- Does not allow underscores in patient ids (see https://hl7.org/fhir/R4/datatypes.html#id)
- More consistent handling of parameter presence to allow for boolean and 0 values (paramPresent function)

## Code changes

- Reorganizes validation logic into a util file
- Adds collectDataPatientIds to manage finding patient ids from different parameters sources
- Moves handling of valueReference.reference to serviceUtils to make parameters consistent between GET and POST
- Expands validateCollectDataParams to validate recognized, supported, required, and single cardinality parameters for $collect-data according to https://hl7.org/fhir/uv/deqm/2026May/en/OperationDefinition-collect-data.html and checks for availability of expected resources (Patient or Group) on the server. Also enforces exclusivity of subject vs subjectGroup which is anticipated to be clarified in the spec (see https://jira.hl7.org/browse/FHIR-57066 ).

# Testing guidance

- `npm run check`
- `npm run start`
- POST to http://localhost:3000/Measure/$collect-data using Insomnia. Example Parameters with subjectGroup below, but can vary parameters to check validation
```
{
"resourceType": "Parameters",
"parameter": [
   {
   	"name": "periodStart",
   	"valueDate": "2023-01-01"
   },
   {
   	"name": "periodEnd",
   	"valueDate": "2023-12-31"
   },
   {
   	"name": "measureUrl",
   	"valueCanonical": "https://madie.cms.gov/Measure/CMS145FHIRCADBetaBlockerTherapyPriorMIorLVSD|0.4.000"
   },
   {
   	"name": "measureUrl",
   	"valueCanonical": "https://madie.cms.gov/Measure/CMS125FHIRBreastCancerScreening"
   },
   {
   	"name": "subjectGroup",
   	"resource": {
			"resourceType": "Group",
			"type": "person",
			"actual": "true",
			"member": [
				{
					"entity": {"reference": "Patient/f6dafa7e-15a9-4fec-baea-59d43201caf8"}
				},
				{
					"entity": {"reference": "Patient/c58acff5-248b-49c9-b18d-69e4a84a08d9"}
				}
			]
		}
   }
  ]
}
```

### DIFF
--- a/src/util/validationUtils.js
+++ b/src/util/validationUtils.js
@@ -312,6 +312,15 @@ function validateCollectDataParams(parameters, reply) {
     );
     return false;
   }
+  if (hasSubjectGroup && parameters.subjectGroup.resourceType !== 'Group') {
+    reply.code(400).send(
+      createOperationOutcome('Parameter subjectGroup must be a resource of type Group.', {
+        issueCode: 400,
+        severity: 'error'
+      })
+    );
+    return false;
+  }
 
   let unsupportedParams = [];
   Object.keys(parameters).forEach(param => {
--- a/src/util/validationUtils.js
+++ b/src/util/validationUtils.js
@@ -1,0 +1,354 @@
+const { findResourceById } = require('./mongo.controller');
+const supportedResources = require('./supportedResources').filter(r => r !== 'ValueSet'); //exclude ValueSet (may be stored but not exported)
+const { createOperationOutcome } = require('./errorUtils');
+
+// Preferable over truthy check to allow for boolean or 0
+function paramPresent(parameters, param) {
+  const value = parameters[param];
+  return value !== undefined && value !== null && value !== '';
+}
+
+// Accept bulkSubmitStatusEndpoint, but currently do nothing with it
+const EXPORT_RECOGNIZED_PARAMS = [
+  '_outputFormat',
+  '_type',
+  '_typeFilter',
+  'patient',
+  '_elements',
+  'organizeOutputBy',
+  'bulkSubmitEndpoint',
+  'bulkSubmitStatusEndpoint'
+];
+
+const COLLECT_DATA_RECOGNIZED_PARAMS = [
+  'measureUrl',
+  'periodStart',
+  'periodEnd',
+  'subject',
+  'subjectGroup',
+  'reporter',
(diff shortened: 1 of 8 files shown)


src/util/validationUtils.js
```diff
@@ -0,0 +1,363 @@
...
+        .send(
+          createOperationOutcome(
+            `The following resourceTypes are not supported for _element param for $export: ${unsupportedResourceTypes.join(
+              ', '
+            )}.`,
+            { issueCode: 400, severity: 'error' }
+          )
+        );
+      return false;
+    } else if (unsupportedElementTypes.length > 0) {
```
elsaperelli, line 171: Ah this was moved over from `export.service.js` but I think this validation still needs to be added? If it's an easy add maybe we can in this PR? If not we should probably track this in a future task :)

elsaperelli (approved):
Just one small comment but not necessary for merging so I will approve!! Nice!!!

src/util/validationUtils.js
lmd59: What do you mean it needs to be added?

lmd59: Looked back at this, and I see what you mean. We don't populate unsupportedElementTypes in any way. I looked back at the [PR](https://github.com/projecttacoma/bulk-export-server/pull/39) in which this was originally added. It looks like there was some thoughtful conversation there about mandatory elements but not so much about unsupported elements. We eventually turn the elements into database projections, so I'm not actually sure what unsupported element types might be. We're fairly flexible in our support for elements. Based on what I see, I would maybe just remove this check entirely.

elsaperelli: @lmd59 and I decided to leave this as is right now and create a task 👍

elsaperelli (approved):
🥳

